PR #1275, opened 2021-11-30
update "eq" filter operator on strapi & strapi-graphql data providers

### LINKED ISSUE #1273: The operator property in initialFilter is always required
**Describe the bug**
When using filters in initialFilter the operator property is mandatory and does not have the option to be omitted.

**To Reproduce**
1.- Create a useTable
2.- Assign object initialFilter 
3.- Add object { field: "modelField", value: "value"}
4.- We get an error: "The property 'operator' is missing in the type '{ field: string; value: string; }' but it is necessary in the type 'CrudFilter'".

**Expected behavior**
The operator property could be optional and not add extra arguments to the url

omeraplak:
Strapi REST and GraphQL APIs don't need a suffix for the "eq" filter. With this change, we can now support Strapi Draft Content filters.

**Closing issues**

closes #1273

### DIFF
--- a/examples/dataProvider/strapi/package.json
+++ b/examples/dataProvider/strapi/package.json
@@ -6,21 +6,21 @@
     "@pankod/refine": "^2.0.7",
     "@pankod/refine-react-router": "^2.0.7",
     "@pankod/refine-strapi": "^2.0.7",
-    "@testing-library/jest-dom": "^5.11.4",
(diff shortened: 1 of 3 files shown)


miyavsu-limited (approved):

salihozdemir (approved):

hectoryx (commented):
Good job

PR #1374, opened 2021-12-17
Change flag for controlling useTable's loading property to isFetched

### LINKED ISSUE #1321: [BUG] Caching does not work with useTable due to the returned loading attribute set to true during re-validation
**Describe the bug**
With the changes introduced in https://github.com/pankod/refine/pull/1319 `useTable` properly returns loading as true while loading is being performed on new requests but now also incorrectly returns true while react query revalidates the response.

**To Reproduce**
1. Go to list page
2. Click on a new page through pagination
3. Observe that loader correctly displays
4. Click on previous page
5. Observe that loader appears again even though data is already there (react query merely revalidates the cache at this point)

**Expected behavior**
Loading should not appear when a cached response is ready to be served. This can be achieved by relying either on `status` or on `isSuccess` instead of `isFetching` (which is also true while revalidating cache).

See for solution suggestion https://github.com/pankod/refine/pull/1319#issuecomment-993494500

**Screenshots**
Also shown in suggestion mentioned above.

**Desktop (please complete the following information):**
 - Browser: Edge
 - Version: 2.4.5
 - Used dataProvider: Custom Mock Implementation

**Additional context**
I wasn't sure if I should create a new issue for this but thought it's better to keep track of it separately since it deals with something slightly different from https://github.com/pankod/refine/issues/1316.

IgnusG:
Changes the `isFetching` flag to the inverted `isFetched` flag instead for setting the loading property. `loading` will thus be true when fetching new data, but set to false if the data is already cached (and react-query is merely re-fetching in the background to update the cache).

See above

**Test plan (required)**
1. Open App
2. Observe initial loading animation
3. Click on a new page (second page) - observe loading animation appear
4. Click on the previous page (first page) - no loading animation (cached result shown, react-query refetches in the background)
5. Click on another new page (third page)
6. Observe the loading animation appear again

Resolves #1321

### DIFF
--- a/packages/core/src/hooks/table/useTable/useTable.ts
+++ b/packages/core/src/hooks/table/useTable/useTable.ts
@@ -193,7 +193,7 @@ export const useTable = <
         liveParams,
         onLiveEvent,
     });
-    const { data, isFetching, isLoading } = queryResult;
+    const { data, isFetched, isLoading } = queryResult;
 
     const onChange = (
         pagination: TablePaginationConfig,
@@ -240,7 +240,7 @@ export const useTable = <
         tableProps: {
             ...tablePropsSunflower,
             dataSource: data?.data,
-            loading: liveMode === "auto" ? isLoading : isFetching,
+            loading: liveMode === "auto" ? isLoading : !isFetched,
             onChange,
             pagination: {
                 ...tablePropsSunflower.pagination,


omeraplak (approved):

omeraplak:
Hey @IgnusG ,
I really liked this. Thanks a lot!

Issue #1373, opened 2021-12-17
[FEAT] Draft: Add support to switch filter operator dynamically from useTable


IgnusG:
**Is your feature request related to a problem? Please describe.**
Currently it is only possible to set the filter operator on the initial call to `useTable`. This suggestion would allow switching between full string filter (eq), negation (ne) or partial string match/not match (contains/ncontains).

Operators are set here initially:
https://github.com/pankod/refine/blob/445baad768c51fc0c734ddf179402de2664ef471/packages/core/src/hooks/table/useTable/useTable.ts#L144-L146

The filter value can be changed but not the operator:
https://github.com/pankod/refine/blob/445baad768c51fc0c734ddf179402de2664ef471/packages/core/src/hooks/table/useTable/useTable.ts#L198-L211

**Describe the solution you'd like**
- Allow <FilterDropdown> changing the operator through props
  - Breaking change?
  - Would antd even allow this for filter?

    ```tsx
    <Table.Column 
        dataIndex="contentName"
        filterDropdown={({ dropdownProps, filterOperatorProps }) => { // Would something like this even work?
            return (
                <Fragment>
                    <Select onChange={filterOperatorProps.onChange}>
                        <Select.Option value="eq">Equals</Select.Option>
                        <Select.Option value="ne">Not Equal</Select.Option>
                    </Select>
                    <FilterDropdown {...dropdownProps}><Input /></FilterDropdown>
                </Fragment>
            )
        }}
    />
    ``` 

- Export a helper method from `useTable`, something like `changeFilterOperator(forField: string, newOperator: CrudOperators)` which could be hooked up manually
  - More work for the developer to hook it up but not breaking
  - Supports changing the filter outside the dropdown too which is cool
  - The operator change input field/select would most likely already be inside of the `dropdownFilter` render prop of a column so specifying the column name/field again seems redundant
 
    ```tsx
     const { changeFilterOperator, ...tableProps } = useTable({});

     ...

     useEffect(() => {
         changeFilterOperator("contentName", "contains"); // We can change the filter operator outside of the dropdown
         // Maybe even the value itself with a slightly different signature
         // changeFilter("contentName", "contains", someWeirdValue);
     }, [someWeirdValue])

    <Table.Column 
        dataIndex="contentName" // First column specification
        filterDropdown={(props) => {
            return (
                <Fragment>
                    <Select onChange={(newOp) => changeFilterOperator("contentName", newOp)}> // "contentName" must be specified here again
[...]

PR #1375, opened 2021-12-17
WIP: Add a new method to useTable for changing the filter operator dynamic…

IgnusG:
This is a prototype of what a simplistic solution to #1373 could look like

This allows users to dynamically change the filter operator - for example a checkbox added inside of `<Table.Column filterDropdown={...} />` could invert the operator from equals (`eq`) to not equal (`ne`), or it could trigger partial matching `contains`/`ncontains` instead of `eq`.

**Test plan (required)**
1. Use the `changeFilterOperator` helper to modify an operator for an existing filter dynamically
2. Observe the operator changing and the data on the page updating

Related to #1373

### DIFF
--- a/packages/core/src/hooks/table/useTable/useTable.ts
+++ b/packages/core/src/hooks/table/useTable/useTable.ts
@@ -30,6 +30,7 @@ import {
     BaseRecord,
     CrudFilters,
     CrudSorting,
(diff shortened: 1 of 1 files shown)


dzcpy:
Any updates?

PR #1408, opened 2022-01-03
Headless Notification with notificationProvider

salihozdemir:
refine headless notification


packages/core/src/hooks/data/useCreate.ts
```diff
@@ -92,11 +93,10 @@ export const useCreate = <
...
                     type: "success",
-                });
-
-                getAllQueries(resource).forEach((query) => {
-                    queryClient.invalidateQueries(query.queryKey);
-                });
+                }),
+                    getAllQueries(resource).forEach((query) => {
+                        queryClient.invalidateQueries(query.queryKey);
+                    });
```
omeraplak, line 99: something seems wrong here?

omeraplak (changes requested):

packages/core/src/hooks/data/useCreate.ts
salihozdemir: absolutely yes, just the forgotten comma at the top of the line. I fixed it.

PR #1486, opened 2022-01-27
WIP : deep strapi-v4 data normalization closes #1485

### LINKED ISSUE #1485: [FIX] strapi-v4 - data provider - deep data normalization
Populated data from strapi-v4 data  provider are not normalized, which makes hard to work with them via ts interfaces. 

Use deep traversal like the one proposed here : https://forum.strapi.io/t/discussion-regarding-the-complex-response-structure-for-rest-graphql-developer-experience/13400/9
```js
const normalize = (data) => {
  const isObject = (data) =>
    Object.prototype.toString.call(data) === '[object Object]'
  const isArray = (data) =>
    Object.prototype.toString.call(data) === '[object Array]'

  const flatten = (data) => {
    if (!data.attributes) return data

    return {
      id: data.id,
      ...data.attributes
    }
  }

  if (isArray(data)) {
    return data.map((item) => normalize(item))
  }

  if (isObject(data)) {
    if (isArray(data.data)) {
      data = [...data.data]
    } else if (isObject(data.data)) {
      data = flatten({ ...data.data })
    } else if (data.data === null) {
      data = null
    } else {
      data = flatten(data)
    }

    for (const key in data) {
      data[key] = normalize(data[key])
    }

    return data
  }

  return data
}
```

Alternative would be to normalize data after you get query data. This might be too late if you load it as useTable which expect slightly modified data already.

**Additional context**
Read more at  : https://forum.strapi.io/t/discussion-regarding-the-complex-response-structure-for-rest-graphql-developer-experience/13400

lukassos:
Replacing normalizeData with new internal implementation keeping function interface. 

Original code was flattening only one layer of data[].attributes, proposed code goes recursively into nested structure. 
This pull req. solves issue : [ #1485](https://github.com/pankod/refine/issues/1485)

To make correct normalization of data we need to make deep inspection of objects and flatten them all. Populated data keep the data/attributes structure strapi-v4 introduced. This is inconsistent with TS interfaces.

**Test plan (required)**
Code was alpha tested, some other unit tests might be neededin scope of Refine project.

example input : https://jsoneditoronline.org/#left=cloud.b7bc52c573a744fd82e0d051b5ad7e5b
example output : https://jsoneditoronline.org/#left=cloud.6ef201011b6849afb749ff552c7095e8

**Closing issues**
#1485

### DIFF
--- a/packages/strapi-v4/src/dataProvider.ts
+++ b/packages/strapi-v4/src/dataProvider.ts
@@ -170,7 +170,7 @@ export const DataProvider = (
     },
 
     create: async ({ resource, variables }) => {
-        const url = `${apiUrl}/${resource}/${id}`;
+        const url = `${apiUrl}/${resource}`;
     
         let dataVariables = { data: variables };
     
--- a/packages/strapi-v4/src/dataProvider.ts
+++ b/packages/strapi-v4/src/dataProvider.ts
@@ -78,12 +78,42 @@ const generateFilter = (filters?: CrudFilters) => {
 };
 
 const normalizeData = (data: any) => {
-    const _data = data.data.map((item: any) => ({
-        id: item.id,
-        ...item.attributes,
-    }));
-
-    return _data;
+    // following code credits goes to roelbeerens see: https://forum.strapi.io/t/discussion-regarding-the-complex-response-structure-for-rest-graphql-developer-experience/13400/9
+    const isObject = (data:any) =>
+      Object.prototype.toString.call(data) === "[object Object]";
+    
+    const flatten = (data:any) => {
+      if (!data.attributes) return data;
+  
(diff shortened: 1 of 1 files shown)


packages/strapi-v4/src/dataProvider.ts
```diff
@@ -140,53 +170,53 @@ export const DataProvider = (
     },
 
     create: async ({ resource, variables }) => {
-        const url = `${apiUrl}/${resource}`;
-
+        const url = `${apiUrl}/${resource}/${id}`;
```
omeraplak, line 173:
```suggestion
        const url = `${apiUrl}/${resource}`;
```
This part was accidentally merged today. Then fixed in master :)

PR #1616, opened 2022-03-08
feat: audit log provider

salihozdemir:
Added Audit Log provider. We will update the documents as soon as possible.


omeraplak (approved):

packages/core/src/components/containers/refine/index.tsx
```diff
@@ -49,6 +51,7 @@ export interface RefineProps {
     routerProvider: IRouterProvider;
     notificationProvider?: NotificationProvider;
     accessControlProvider?: AccessControlProvider;
+    auditLogProvider?: IAuditLogContext;
```
salihozdemir, line 54:
```suggestion
    auditLogProvider?: AuditLogProvider;
```
I think, we should use it with the type in `packages/core/src/contexts/auditLog/IAuditLogContext.ts` file.

packages/core/src/contexts/auditLog/index.tsx
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+
+import { IAuditLogContext } from "./IAuditLogContext";
+
+/** @deprecated default value for access control context has no use and is an empty object. */
+export const defaultAuditLogContext: IAuditLogContext = {};
```
salihozdemir, line 6:
```suggestion

```
I think, we should delete them. Because we didin't export `defaultAuditLogContext` before. So we don't need to mark it to deprecated.

salihozdemir:
Are we going to export `antd-audit-log` package?

packages/antd-audit-log/package.json
```diff
@@ -0,0 +1,60 @@
+{
+    "name": "@pankod/refine-antd-audit-log",
+    "description": "refine audit log component for antd package. refine is a React-based framework for building internal tools, rapidly. It ships with Ant Design System, an enterprise-level UI toolkit.",
+    "version": "3.8.0",
+    "license": "MIT",
+    "main": "dist/index.js",
+    "typings": "dist/index.d.ts",
+    "private": false,
```
omeraplak, line 8:
```suggestion
    "private": true,
```
we can mark as a private for disable npm publish

yildirayunlu (approved):

PR #1641, opened 2022-03-14
Make OptionsProps generic

Bastian:
This allows the user to add a type-definition to the `IResourceComponentsProps` for the `options` prop instead of defaulting to `any`.
Example usage:

```tsx
<Refine
    resources=[{
        name: "posts",
        options: {
            dataProviderName: "my-custom-dataprovider",
        }
    }]
    ...
/>
```
```ts
export const PostList: React.FC<IResourceComponentsProps<{ dataProviderName: string }>> = (props) => {
    const { tableProps, searchFormProps, filters, ...rest } = useTable<IPost>({
        dataProviderName: props.options?.dataProviderName
    });
    // ...
}
```

If no generic param is provided, it default to the previous `[key: string]: any` definition.

### DIFF
--- a/packages/core/src/contexts/resource/IResourceContext.ts
+++ b/packages/core/src/contexts/resource/IResourceContext.ts
@@ -3,11 +3,10 @@ import { ReactNode } from "react";
 export interface IResourceContext {
     resources: IResourceItem[];
 }
-export interface OptionsProps {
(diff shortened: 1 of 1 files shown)


omeraplak (approved):
Wow, looks so cool! Thank you 🚀

PR #1668, opened 2022-03-30
Fix correction of example code for components catchAll

nrikiji:
Warning: Functions are not valid as a React child. This may happen if you return a Component instead of <Component /> from render. Or maybe you meant to call this function rather than return it.

### DIFF
--- a/documentation/docs/core/components/refine-config.md
+++ b/documentation/docs/core/components/refine-config.md
@@ -325,7 +325,7 @@ const App: React.FC = () => {
         <Refine
             ...
             // highlight-next-line
-            catchAll={CustomErrorPage}
+            catchAll={CustomErrorPage()}
         />
     );
 };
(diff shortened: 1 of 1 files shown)


biskuvit (approved):

omeraplak:
Hey @nrikiji ,
Congratulations and many thanks. Wouldn't it be better if we change `CustomErrorPage` from `React.FC` to `React.Element`? What do you say?

like this,
```
const CustomErrorPage = <div>Page not found</div>;

const App: React.FC = () => {
    return (
        <Refine
            ...
            catchAll={CustomErrorPage}
        />
    );
};
```

nrikiji:
```
const CustomErrorPage = <div>Page not found</div>;

const App: React.FC = () => {
    return (
        <Refine
            ...
            catchAll={CustomErrorPage}
        />
    );
};
```
Thank you. This description is very clear.

PR #1684, opened 2022-04-03
Updated docs regarding setting Authorization Credentials

### LINKED ISSUE #1683: [DOC] Update "Setting Authorization Credentials"
**Is this issue related to an existing part of the documentation? If so, please provide a link**
The authorization credential part of the docs is a little vague. I think it would be useful for beginners to have more complete documentation. 

[Setting Authorization Credentials](https://refine.dev/docs/core/providers/auth-provider/#setting-authorization-credentials)

**Describe the thing to improve**
The importance of a singular Axios instance and other methods for setting credentials.

IanSaucy:
First version of updated docs for the `Authorization Credentials` section of `core/providers/auth-provider`. 

I added some language about the importance of a singular Axios instance and another method for setting credentials. 

I also fixed a typo in the original example. Switching from `axiosInstance.defaults.headers` -> ` axiosInstance.defaults.headers.common`.

One possible issue is that the example interceptor code is synchronous and Axios has some special docs about this. I was unable to get their example working though. See [here](https://github.com/axios/axios#interceptors) for more information.

I also  think it could be useful to provide an example for something that is _not_ Bearer style authorization. I'm not sure what the next most common authorization format would be -- but if a suggestion is given I could write some example docs for that type as well. 

I know this PR will need some work, just want to get an Alpha version going to start the conversation and see how we can make this page better (: 

Closes #1683

### DIFF
--- a/documentation/docs/core/providers/auth-provider.md
+++ b/documentation/docs/core/providers/auth-provider.md
@@ -527,7 +527,7 @@ If the resolved data has a `name` property, a name text appears; if it has an `a
 
 After user logs in, their credentials can be sent along with the API request by configuring the [`dataProvider`](/core/providers/data-provider.md). A custom `httpClient` can be passed to [`dataProvider`](/core/providers/data-provider.md) to include configurations like cookies and request headers.
 
-We'll show how to add a token acquired from the `login` method to the **Authorization** header of the **HTTP** requests.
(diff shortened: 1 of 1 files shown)


omeraplak (approved):
Very good! Congrats. It's much more descriptive. I think the Token example is sufficient because in general API authentications work in similar ways

PR #1710, opened 2022-04-11
WIP: Make links more accessible.

rassie:
This is a work-in-progress PR which intends to fix #1709. 

The first iteration of this PR only includes changes to `<CloneButton />`, while still adding all necessary machinery for other buttons. It is meant for a first discussion on implementation and code style before applying the same style to the other buttons.

**Problems**
I do not like the way `react-router` recommends testing `location` objects with `render` property (which is what this patch is based on). It feels unnecessarily verbose und complex, but I have yet to see a method that works at all and looks better. On the positive side, this kind of testing actually tests the result instead of implementation -- previously the test checked whether a function has been called, while not verifying that the end result is what is expected.

**Work to do**
A lot of other components need to be updated apart from the buttons:

- [ ] Sider
- [ ] SimpleTable
- [ ] etc.

PR #1835, opened 2022-05-25
Parse query string even without syncWithLocation

rassie:
Follow-up to #1710. 

This makes sure that hotlinking will work even if `syncWithLocation` is not used.

### DIFF
--- a/packages/core/src/hooks/useTable/index.ts
+++ b/packages/core/src/hooks/useTable/index.ts
@@ -112,15 +112,15 @@ export const useTable = <
     let defaultSorter = initialSorter;
     let defaultFilter = initialFilter;
 
-    if (syncWithLocation) {
-        const { parsedCurrent, parsedPageSize, parsedSorter, parsedFilters } =
-            parseTableParams(search);
+    // We want to always parse the query string even when syncWithLocation is
+    // deactivated, for hotlinking to work properly
+    const { parsedCurrent, parsedPageSize, parsedSorter, parsedFilters } =
+        parseTableParams(search);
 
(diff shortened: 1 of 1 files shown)


packages/core/src/hooks/useTable/index.ts
```diff
@@ -155,13 +155,13 @@ export const useTable = <
     };
 
     useEffect(() => {
-        if (syncWithLocation && search === "") {
+        if (search === "") {
```
omeraplak, line 158: I am not sure about this line, i will do a debug

omeraplak (commented):
Thank you!

packages/core/src/hooks/useTable/index.ts
omeraplak: https://github.com/pankod/refine/issues/1810 
We added it for this issue, it made sense to me at the time. What do you think?

omeraplak: this line works as expected no problem 👍 

But I found another error:
If `syncWithLocation` is `false` and I do a filter the URL stays the same. Actually, we expect the current to be `1`. How do you think we can handle this situation?

https://user-images.githubusercontent.com/1110414/170302982-9ecbd02d-0db6-4c76-9ac8-75d63011807c.mov

omeraplak (approved):

packages/core/src/hooks/useTable/index.ts
omeraplak: I guess it's not a bug after`syncWithLocation` is turned off and maybe requested not to be tracked :)

PR #1873, opened 2022-06-01
refactor: remove default values from refine contexts

aliemir:
Removed dummy default values from internal contexts.
Updated contexts:

-   Auth
-   Access Control
-   Notification
-   Translation (i18n)
-   unsavedWarn

### DIFF
--- a/examples/fineFoods/admin/src/components/header/index.tsx
+++ b/examples/fineFoods/admin/src/components/header/index.tsx
@@ -167,7 +167,7 @@ export const Header: React.FC = () => {
     }, [value]);
 
     const menu = (
-        <Menu selectedKeys={[currentLocale]}>
+        <Menu selectedKeys={currentLocale ? [currentLocale] : []}>
             {[...(i18n.languages ?? [])].sort().map((lang: string) => (
                 <Menu.Item
                     key={lang}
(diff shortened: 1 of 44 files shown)


packages/antd/src/components/crud/edit/index.spec.tsx
```diff
@@ -1,14 +1,16 @@
...
 
 import { render, TestWrapper, waitFor } from "@test";
 import { Edit } from "./";
 
 const renderEdit = (
     edit: ReactNode,
-    accessControlProvider?: AccessControlProvider,
+    accessControlProvider?: ComponentProps<
+        typeof Refine
+    >["accessControlProvider"],
```
salihozdemir, line 13:
```suggestion
    accessControlProvider?: AccessControlProvider
```

I think it would be better if you wrote it like this.

aliemir: updated the usage as suggested

examples/fineFoods/admin/src/components/header/index.tsx
```diff
@@ -167,7 +167,7 @@ export const Header: React.FC = () => {
     }, [value]);
 
     const menu = (
-        <Menu selectedKeys={[currentLocale]}>
+        <Menu selectedKeys={currentLocale ? [currentLocale] : []}>
```
omeraplak, line 170: looks cool 👍 

we should update superplate too,

https://github.com/pankod/superplate-core-plugins/blob/master/refine-react/plugins/i18n-antd/src/components/layout/header/index.tsx#L30

https://github.com/pankod/superplate-core-plugins/blob/master/refine-nextjs/plugins/i18n-antd/src/components/layout/header/index.tsx#L25

aliemir: Created a PR with respective changes. https://github.com/pankod/superplate-core-plugins/pull/267 🥂

PR #1878, opened 2022-06-02
refactor(buttons): svgIconProps passed to buttons

omeraplak:
In this PR, We have been passed svgIconProps to all buttons that served from @pankod/refine-mui.


.changeset/polite-impalas-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pankod/refine-mui": minor
```
omeraplak, line 2:
```suggestion
"@pankod/refine-mui": patch
```

PR #1888, opened 2022-06-06
feat: extend `useTable.setFilters` function

aliemir:
Currently, `setFilters` function returned from the `useTable` hook (package: `@pankod/refine-core`) works with `merge` behavior. meaning; when you pass an array of `CrudFilter`s, it will append the new filters to the existing ones and remove the duplicates. 

We're introducing a second parameter to the `setFilters` function: `behavior`. It's default to `merge` but can also be `replace` which might be the preferred behavior for some use cases. 

**Usage**
```ts
setFilters(newFilters: CrudFilters, behavior: "merge" | "replace")
```

Default behavior can be set through `useTable` parameters with `defaultSetFilterBehavior`

**Usage**
```ts
const { /* */ } = useTable({
  /* ... */
  defaultSetFilterBehavior: "replace",
  /* ... */
}); 
```

We're also including an overload for the `setFilters` function as a setter function. 
**Usage**
```ts
setFilters((prevFilters: CrudFilters) => CrudFilters)
```

**Status**

- [x] feat: `merge` and `replace` behaviors added
- [x] feat: setter function overload added
- [x] test: add new test cases for the new behaviors
- [x] bug: remove filter is not working properly on `merge`
- [x] docs: `useTable` docs needs to be updated


.changeset/forty-humans-carry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pankod/refine-mui": patch
```
aliemir, line 2:
```suggestion
"@pankod/refine-mui": minor
```

.changeset/moody-pandas-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pankod/refine-antd": patch
```
aliemir, line 2:
```suggestion
"@pankod/refine-antd": minor
```

documentation/docs/core/hooks/useTable.md
```diff
@@ -154,28 +158,73 @@ setFilter([
 ]);
 
 console.log(filters); // [{ field: "title", operator: "contains", value: "A" }, { field: "status", operator: "equals", value: "published" }]
+
+setFilter([
+    {
+        field: "author",
+        operator: "contains",
+        value: "Ali",
```
omeraplak, line 166:
```suggestion
        value: "Foo",
```

packages/antd/src/hooks/table/useTable/useTable.ts
```diff
@@ -97,6 +98,7 @@ export const useTable = <
         initialCurrent,
         initialPageSize,
         initialSorter,
+        defaultSetFilterBehavior,
```
salihozdemir, line 101: `useTable` hook doesn't export the `setFilters`, `setSorters` etc. in `@pankod/refine-antd` package so `defaultSetFilterBehavior` doesn't create a use case  🤔

aliemir, line 101:
```suggestion
```

salihozdemir (approved):

PR #1896, opened 2022-06-08
Feat: lazy notification handle calls

aliemir:
## `@pankod/refine-core`

Added an overload to `notificationProvider` in `Refine` wrapper component from `@pankod/refine-core`. 

Basically, we may need this if the `notificationProviderHandle` is using some hooks that are also being provided in the same component as the `Refine` component. Instead of creating couple components to wrap those up and have a complexity; we can now just lazily initialize the notification handler for **refine** to use.

**Old Way (Still supported with no changes)**
```ts
/* ... */
const notificationProvider = notificationProviderHandle();

return (
  <Refine
      notificationProvider={notificationProvider}
      /* ... */
  />
);
```
**Overloaded Way**
```ts
/* ... */

return (
  <Refine
      notificationProvider={notificationProviderHandle}
      /* ... */
  />
);
```

## `@pankod/refine-mui`

Added missing exports for `notistack` to `@pankod/refine-mui` package.


omeraplak (approved):
🎖️

ozkalai (approved):

yildirayunlu (approved):

salihozdemir (approved):

PR #1910, opened 2022-06-09
fix: `useDatagrid` filter bug

salihozdemir:
Fixed the `useDatagrid` filter bug that the selected filter was not displayed.

### DIFF
--- a/.changeset/large-shrimps-begin.md
+++ b/.changeset/large-shrimps-begin.md
@@ -2,4 +2,4 @@
 "refinejs-repo": patch
 ---
 
-Fixed the `useDatagrid` filter bug that the selected filter was not displayed.
+Fixed the `useDataGrid` filter bug that the selected filter was not displayed.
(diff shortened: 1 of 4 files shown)


.changeset/large-shrimps-begin.md
```diff
@@ -0,0 +1,5 @@
+---
+"refinejs-repo": patch
+---
+
+Fixed the `useDatagrid` filter bug that the selected filter was not displayed.
```
aliemir, line 5:
```suggestion
Fixed the `useDataGrid` filter bug that the selected filter was not displayed.
```

packages/mui/src/definitions/dataGrid/index.ts
```diff
@@ -74,7 +74,7 @@ export const transformFilterModelToCrudFilters = ({
 }: GridFilterModel): CrudFilters => {
     const filters = items.map(({ columnField, value, operatorValue }) => ({
         field: columnField,
-        value,
+        value: value ?? "",
```
aliemir, line 77: With this change, I guess we can't have `undefined` as a value now, right? Is this going to be an issue on removing filters in `merge` mode? 🤔 @salihozdemir

PR #1985, opened 2022-06-17
test: update `useMenu` tests

aliemir:
While updating `useMenu` tests, I've noticed that `<Sider/>` in `@pankod/refine-antd` uses the same key for all items and resulting in opening all menus at once initially.

- [x] updated `useMenu` tests in `@pankod/refine-core`
- [x] updated `Menu.Item.keys` in `@pankod/refine-antd`'s `<Sider/>`
- [x] updated `<Sider />` in documentations and examples
- [x] updated `TestWrapper` in `@pankod/refine-core` with resources modifiers.
- [x] reverted behavioral change in _deprecated_ `useMenu` hook in `@pankod/refine-antd`

### DIFF
--- a/.changeset/sweet-tigers-pay.md
+++ b/.changeset/sweet-tigers-pay.md
@@ -1,0 +1,5 @@
+---
+"@pankod/refine-antd": patch
+---
+
+Update `key`s in `<Sider/>` component to use `route`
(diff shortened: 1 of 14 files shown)


aliemir:
Also applied the changes here https://github.com/pankod/superplate-core-plugins/pull/269

omeraplak (approved):

PR #2021, opened 2022-06-22
feat: new link behavior in nextjs router

aliemir:
Added `legacyBehavior: false` to `defaultProps` to `<Link/>` component in `@pankod/refine-nextjs-router`

### DIFF
--- a/.changeset/smart-horses-cheer.md
+++ b/.changeset/smart-horses-cheer.md
@@ -1,0 +1,5 @@
+---
+"@pankod/refine-nextjs-router": patch
+---
+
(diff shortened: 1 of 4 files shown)


salihozdemir (approved):

omeraplak (approved):

aliemir:
@omeraplak @salihozdemir @biskuvit @ozkalai I added one more check to `RefineLink` export. The type is `function` but in runtime imports it's an `object` and `bind` was causing an error 🤷‍♂️ I needed to add a condition for that with `@ts-expect-error`

ozkalai (approved):

PR #2026, opened 2022-06-22
chore(base-headless-example): moved under the base examples

ozkalai:
In this PR, We have moved `baseHeadless` example under the `base` example folder.


aliemir (approved):

aliemir:
Maybe we can rename this directory to `headless` instead of `baseHeadless`? 🤔

PR #2030, opened 2022-06-22
feat(core): add translate support to useBreadcrumb

biskuvit:
add translate support to [`useBreadcrumb`](https://refine.dev/docs/core/hooks/useBreadcrumb/)

### DIFF
--- a/.changeset/calm-sloths-pretend.md
+++ b/.changeset/calm-sloths-pretend.md
@@ -1,0 +1,5 @@
+---
+"@pankod/refine-core": minor
+---
+
+add translate support to [`useBreadcrumb`](https://refine.dev/docs/core/hooks/useBreadcrumb/)
--- a/.changeset/calm-sloths-pretend.md
+++ b/.changeset/calm-sloths-pretend.md
@@ -2,4 +2,4 @@
 "@pankod/refine-core": minor
 ---
 
-add translate support to [`useBreadcrumb`](https://refine.dev/docs/core/hooks/useBreadcrumb/)
+Added i18n support for resource names on [`useBreadcrumb`](https://refine.dev/docs/core/hooks/useBreadcrumb/) hook.
--- a/packages/core/src/hooks/breadcrumb/index.ts
+++ b/packages/core/src/hooks/breadcrumb/index.ts
@@ -71,7 +71,7 @@ export const useBreadcrumb = (): UseBreadcrumbReturnType => {
 
     if (action) {
         breadcrumbs.push({
-            label: translate(`${action}`, humanizeString(action)),
+            label: humanizeString(action),
         });
     }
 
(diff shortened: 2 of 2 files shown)


packages/core/src/hooks/breadcrumb/index.ts
```diff
@@ -52,13 +59,20 @@ export const useBreadcrumb = (): UseBreadcrumbReturnType => {
...
+                humanizeString(resource.name),
+            ),
         href: !!resource.list ? `/${resource.route}` : undefined,
         icon: resource.icon,
     });
 
     if (action) {
-        breadcrumbs.push({ label: humanizeString(action) });
+        breadcrumbs.push({
+            label: translate(`${action}`, humanizeString(action)),
```
omeraplak, line 74:
```suggestion
            label: translate(`${action}`, humanizeString(action)),
```
Do we have a translate for this key? maybe `buttons.${action}` can be used https://refine.dev/docs/core/providers/i18n-provider/#adding-translations-files

@salihozdemir @aliemir

.changeset/calm-sloths-pretend.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pankod/refine-core": minor
+---
+
+add translate support to [`useBreadcrumb`](https://refine.dev/docs/core/hooks/useBreadcrumb/)
```
omeraplak, line 5:
```suggestion
Added i18n support for resource names on [`useBreadcrumb`](https://refine.dev/docs/core/hooks/useBreadcrumb/) hook
```

packages/core/src/hooks/breadcrumb/index.ts
aliemir: I think using `buttons` would lead to confusion and unwanted side effects 🤔 

I think we can either
1. have an another scope `breadcrumb` and use the key like `breadcrumb.${action}` 
2. rename `buttons` to something like `actions` and use them both in the buttons and the breadcrumbs
3. This one is my favorite but more complex looking; nest them inside a resource property to give the user an easy way to use different action labels in different resources. Something like `${resource}.actions.${action}`. 

With the 3rd way, for example in Fine Foods examples, we can have `Courier > Add Courier` and `Category > Create Category` labels for `create` action 🤔

salihozdemir: > Maybe we can apply both 2nd and 3rd items from my message above 🤔
> 
> Something like this with 2nd as the fallback of the 3rd:
> 
> ```ts
> translate(`${resource}.actions.${action}`, translate(`actions.${action}`, humanizeString(action)))
> ```

Aren't we complicating it a bit too much?  😅

omeraplak (approved):

aliemir (approved):

PR #2044, opened 2022-06-23
fix: material UI finefoods example bugs

salihozdemir:
Fix Material UI example bugs.


examples/fineFoods/admin/mui/src/components/dashboard/deliveryMap/index.tsx
```diff
@@ -4,6 +4,7 @@ import GoogleMapReact from "google-map-react";
 import { MapMarker } from "components/mapMarker";
 import { LocationIcon, CourierIcon } from "components/icons";
 import { IOrder } from "interfaces";
+import { Box } from "@pankod/refine-mui";
```
omeraplak, line 7: Can we move it to the top?

PR #2072, opened 2022-06-29
fix: remove `useDataGrid` `columns` dependency

salihozdemir:
The `useDataGrid` hook required the `columns` property. Since this situation gives us some difficulties, we will remove the column dependency.

### DIFF
--- a/packages/mui/src/hooks/useDataGrid/index.ts
+++ b/packages/mui/src/hooks/useDataGrid/index.ts
@@ -94,10 +94,11 @@ export const useDataGrid = <
     liveParams,
     metaData,
     dataProviderName,
-}: UseDataGridProps<TData, TError, TSearchVariables>): UseDataGridReturnType<
+}: UseDataGridProps<
     TData,
+    TError,
     TSearchVariables
-> => {
+> = {}): UseDataGridReturnType<TData, TSearchVariables> => {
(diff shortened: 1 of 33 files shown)


aliemir:
This PR is mentioned in and planned to resolve the issue https://github.com/pankod/refine/issues/2062

packages/mui/src/hooks/useDataGrid/index.ts
```diff
@@ -249,9 +257,12 @@ export function useDataGrid<
             filterMode: "server",
             filterModel: transformCrudFiltersToFilterModel(
                 differenceWith(filters, permanentFilter ?? [], isEqual),
-                columns,
+                columnsLookup,
             ),
             onFilterModelChange: handleFilterModelChange,
+            onStateChange: (state) => {
+                setColumnsLookup(state.columns.lookup);
```
aliemir, line 264: I'm not sure how is this handled but since the state change is done frequently by `<DataGrid/>` do we need to have some kind of equality check to avoid re-calculations? 🤔 


Maybe we can check if the reference is same or not, if so, we may not need any checks here. 

But if all changes trigger the state change on `columnsLookup` this might degrade the performance on data heavy apps. Maybe we can map the `state.columns.lookup` to have a leaner/simpler data type for us to check and do some deep equality check before setting the state 🤔

omeraplak (approved):

omeraplak (approved):

PR #2260, opened 2022-08-01
fix: upgrade react query version to 4

salihozdemir:
Upgraded `react-query` version to 4.

- [x] Fix the `react-hook-form`'s `useModalForm` console errors.

### Test plan (required)

In progress.

### Closing issues

-

### Self Check before Merge


packages/core/src/hooks/auditLog/useLog/index.spec.ts
```diff
@@ -39,7 +39,7 @@ describe("useLog Hook", () => {
                 meta: {
                     id: 1,
                 },
-                author: undefined,
+                author: {},
```
omeraplak, line 42: Do we need to change this line?

salihozdemir: Yes, because `useGetIdentity` return `{}` instead of `undefined`.

omeraplak (approved):

PR #2298, opened 2022-08-08
feat: keep query params on syncWithLocation

omeraplak:

Fixed user-defined URL query parameters being deleted when using `syncWithLocation`

### Self Check before Merge


packages/core/src/hooks/useTable/index.ts
```diff
@@ -180,13 +181,16 @@ export function useTable<
         sorter,
         filters,
     }: SyncWithLocationParams) => {
+        const currentQueryParams = qs.parse(search?.substring(1)); // remove first ? character
```
salihozdemir, line 184: In this line, May we use params returned from `useParams`? Thus, we don't need to `substring`.

packages/core/src/hooks/useTable/index.ts
```diff
@@ -202,6 +206,7 @@ export function useTable<
 
     useEffect(() => {
         if (syncWithLocation) {
+            const currentQueryParams = qs.parse(search?.slice(1)); // remove first ? character
```
salihozdemir, line 209: Same thing the above comment.

packages/core/src/hooks/useTable/index.ts
omeraplak: `useParams` works for URL parameters but does not return query parameters

salihozdemir (approved):

PR #2312, opened 2022-08-10
feat(antd): separate `refine-antd` styles

aliemir:
Separated `styles.min.css` file as `antd.min.css` and `reset.min.css` to make users able to turn off `reset` styles when needed.

If you want to use both of those, you can continue using `styles.min.css` like the following:

```tsx
import "@pankod/refine-antd/dist/styles.min.css";
```

In some cases, users might want to disable `reset` and only apply the `antd` specific styles, in those cases; you can import like the following:

```tsx
// import  "@pankod/refine-antd/dist/reset.min.css"; // `styles.min.css` includes both of these files.
// importing `antd.min.css` will only apply antd specific styles and ignore the reset.
import  "@pankod/refine-antd/dist/antd.min.css";
```

### Test plan (required)

Tests are not affected.

### Self Check before Merge


aliemir:
In my use case, I was working on live preview feature in the docs and the antd styles was conflicting and overriding the global styles of Docusaurus. In this case, I needed to only import antd specific styles and skip the reset part.

aliemir:
We don't have any references on why we do this style import, maybe later we can add a section for this and also mention this change in there.

omeraplak (approved):

ozkalai (approved):

PR #2352, opened 2022-08-23
feat: add `options` prop to `<Refine>`

### LINKED ISSUE #2133: [FEAT] Allowing an `options` prop in `<Refine/>` to configure default values/behaviors
**refine** API has defaults values in its hooks and helpers which can ben overridden from their parameters but in some cases, user might need to apply their overrides at every usage. For example the default redirect in `useForm` which is set to `edit` (was `list` in older versions).

An `options` prop to `<Refine/>` component to edit default values (not all of them but the ones most changed). While doing so, we can also move `mutationMode`, `syncWithLocation`, `warnWhenUnsavedChanges`, `undoableTimeout`, `reactQueryClientConfig` and `reactQueryDevtoolConfig` props to `options` (or `config`) for sake of consistency.

Currently, default values can be overridden in place but this can lead to repetition of the same lines over and over in the whole project.

Still not sure what's the best approach here, waiting for comments from the community. Do you have any overrides you do in every usage of a hook or a component? Do you think having a way to change the defaults might work for your case?

salihozdemir:
Add a new `<Refine>` component property: `options`.

Previously, the options were passed as a property to the `<Refine>`. Now, the options are passed to the `<Refine>` via `options` property like this:

```diff
    <Refine
-       mutationMode="undoable"
-       undoableTimeout={5000}
-       warnWhenUnsavedChanges
-       syncWithLocation
-       liveMode="off"
-       disableTelemetry={false}
+       options={{
+           mutationMode: "undoable",
+           undoableTimeout: 5000,
+           warnWhenUnsavedChanges: true,
+           syncWithLocation: true,
+           liveMode: "off",
+           disableTelemetry: false,
+       }}
    />
```

### Test plan (required)

All tests passed.

### Closing issues

fix #2133 

### Self Check before Merge

### DIFF
--- a/packages/core/src/components/containers/refine/index.tsx
+++ b/packages/core/src/components/containers/refine/index.tsx
@@ -160,7 +160,7 @@ export const Refine: React.FC<RefineProps> = ({
                 },
             },
         });
-    }, [reactQueryClientConfig]);
+    }, [reactQueryClientConfig, options?.reactQuery?.clientConfig]);
 
     const notificationProviderContextValues = React.useMemo(() => {
         return typeof notificationProvider === "function"
--- a/packages/core/src/components/containers/refine/index.tsx
+++ b/packages/core/src/components/containers/refine/index.tsx
@@ -13,7 +13,7 @@ import { DataContextProvider } from "@contexts/data";
 import { LiveContextProvider } from "@contexts/live";
 import { TranslationContextProvider } from "@contexts/translation";
 import { ResourceContextProvider, IResourceItem } from "@contexts/resource";
-import { defaultRefineConfig, RefineContextProvider } from "@contexts/refine";
+import { defaultRefineOptions, RefineContextProvider } from "@contexts/refine";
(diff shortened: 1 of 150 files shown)


packages/core/src/components/containers/refine/index.tsx
```diff
@@ -61,24 +61,46 @@ export interface RefineProps {
     LoginPage?: React.FC;
     DashboardPage?: React.FC;
     ReadyPage?: React.FC;
+    /** @deprecated mutationMode is deprecated. Use options instead. https://refine.dev/docs/core/components/refine-config/#mutationmode */
```
omeraplak, line 64:
```suggestion
    /** @deprecated 
    "mutationMode" property is deprecated at this level. Use it from within "options" instead.
    options={{ mutationMode="optimistic" }}
    https://refine.dev/docs/core/components/refine-config/#mutationmode */
```

omeraplak (approved):

PR #2384, opened 2022-08-29
test: update async timeouts globally

aliemir:
Fixing the random failing issue of tests on CI.

### Test plan (required)

Still on draft

### Self Check before Merge

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -12,44 +12,44 @@ jobs:
             matrix:
                 node-version: [16.x]
         steps:
-        - uses: actions/checkout@v1
-        - name: Cache node modules
-            uses: actions/cache@v2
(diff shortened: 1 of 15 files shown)


aliemir:
I'll mark this as ready after 6 consecutive successful runs. This was the first.

aliemir:
:(

omeraplak:
> :(

image

aliemir:
I set up an action with 10 jobs and every job runs the tests twice. Currently total consecutive success count is 17. If all succeed, it will be 38.

aliemir:
I'm not super confident to say they won't fail again but at least we know we've fixed some smells and wrong structured tests. There may be two or three more suits might fail but we can fix those when we see them.

omeraplak (approved):

PR #2389, opened 2022-08-30
feat: add option to hide resource in the menu

### LINKED ISSUE #2388: [FEAT] Add an option to hide the resource in the menu

Currently, users need to customize the `<Sider />` components to hide resources from their menu. We should support a way in our `useMenu` hook at `@pankod/refine-core` to let users hide resources in the menu. 

A field on `options` property of the resources.

```tsx
<Refine
    resources={[
        {
            name: "posts",
            options: {
                visible: false, // or hide: true
            }
        }
    ]}
/>
```

alicanerdurmaz:
"hide" property added to the resource options.
With this way we are able to hide menu items in the <Sider/> while rendering.

`<Refine
    resources={[
        {
            name: "posts",
            options: {
                hide: true
            }
        }
    ]}
/>`

I prefer to use **hide** keyword because it seems to me more easy to use and If we use **visible** keyword, we need to provide default value to true. 

`if(hide) return null`
`if(!visible) return null`

### Test plan (required)
All tests are passed.
<p align="center">

</p>

### Closing issues
closes #2388

### Self Check before Merge

### DIFF
--- a/.changeset/breezy-icons-doubt.md
+++ b/.changeset/breezy-icons-doubt.md
@@ -1,0 +1,8 @@
+---
+"@pankod/refine-antd": minor
+"@pankod/refine-core": minor
+"@pankod/refine-mui": minor
+"@pankod/refine-ui-tests": minor
+---
+
+add an option to hide the source in the menu #2388
--- a/documentation/docs/core/hooks/ui/useMenu.md
+++ b/documentation/docs/core/hooks/ui/useMenu.md
@@ -269,6 +269,7 @@ interface IResourceItem extends IResourceComponents {
     canEdit?: boolean;
     canShow?: boolean;
     canDelete?: boolean;
+    options?: OptionsProps;
     parentName?: string;
 }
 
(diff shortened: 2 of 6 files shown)


aliemir:
Hey @alicanerdurmaz thank you for the great work! I had a different approach in mind and an edge case that we need to support, let me explain my approach and maybe we can discuss and proceed;

- Instead of handing the `hide` option in `Sider` components, we can use `useMenu` hook in the `refine-core`. Both `Sider` components use the `useMenu` hook so it will be a single point of change.
- I'm guessing filtering the result of `useResource` hook (returns the array of resources) inside `useMenu` will be enough.
- But actually, we also need to filter sub elements like if we have a parent resource `cms` and it's hidden, we should also hide its child resources.
- This probably was not an issue when we update the `Sider` components like you do, so `useMenu` approach has this one downside.
- I'm guessing `useMenu` will be a better place to handle this since we will also be embracing the headless users.

Looking forward to hear your take 🥂

alicanerdurmaz:
> Hey @alicanerdurmaz thank you for the great work! I had a different approach in mind and an edge case that we need to support, let me explain my approach and maybe we can discuss and proceed;
> 
> * Instead of handing the `hide` option in `Sider` components, we can use `useMenu` hook in the `refine-core`. Both `Sider` components use the `useMenu` hook so it will be a single point of change.
> * I'm guessing filtering the result of `useResource` hook (returns the array of resources) inside `useMenu` will be enough.
> * But actually, we also need to filter sub elements like if we have a parent resource `cms` and it's hidden, we should also hide its child resources.
> * This probably was not an issue when we update the `Sider` components like you do, so `useMenu` approach has this one downside.
> * I'm guessing `useMenu` will be a better place to handle this since we will also be embracing the headless users.
> 
> Looking forward to hear your take 🥂

i like your approach more than mine. Thank you for clear explanation 🙏

PR #2427, opened 2022-09-02
feat: Issue #2426 Add nesting sorting

### LINKED ISSUE #2426: [FEAT] Sort list/table by Nested Field
I am using Hasura/Ant and cannot sort a column based on a value in a nested field.
I  want to be able to sort a column based on the display value which is in a nested field.

Using the compare function in sorter which doesn't seem to work

**Additional context**
Example column data:
```
{
        "id": "b025b2ab-387a-4e03-a902-a4ecc19dd922",
        "name": "example thing",
        "tag": {
          "name": "example tag",
          "value": "A-1" <-- sort by this value
        }
}
```

geoffatsource:
I've split on commas here in order to directly support the `dataIndex` array in table columns. It seems to work for me.

This is related to Issue 2426 [https://github.com/pankod/refine/issues/2426](url) adding the ability to sort based on nested fields used in array type dataIndex values in table columns

### Test plan (required)

This would require creating a list view with nested data as shown in the feature request and adding a sorter flag

### Closing issues

closes #2426

### Self Check before Merge

-   [ x] TypeScript definitions are updated/provided or not needed

### DIFF
--- a/.changeset/fast-rocks-fly.md
+++ b/.changeset/fast-rocks-fly.md
@@ -1,5 +1,6 @@
 ---
 "@pankod/refine-hasura": patch
+"@pankod/refine-nhost": patch
 ---
 
 Added nested sorting feature 💥
--- a/.changeset/fast-rocks-fly.md
+++ b/.changeset/fast-rocks-fly.md
@@ -1,0 +1,5 @@
+---
+"@pankod/refine-hasura": patch
+---
+
+Added nested sorting feature 💥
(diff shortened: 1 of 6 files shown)


ozkalai:
hey @geoffatsource, Thank you for creating this pull request. I couldn't perfectly understand why you use a comma instead of a dot to split sort fields. What do you think about using dot notation like the filter we used instead of a comma, or is there anything I missed?

geoffatsource:
I'd think that's fine, however it appears that `dataIndex` with arrays as values are joined, so they automatically come in with commas. I'd think replacing commas with dots, then splitting would cover both cases. :)

PR #2482, opened 2022-09-12
feat(mantine): add mantine

omeraplak:

We will merge this PR, but we need to make a few updates before the release. Thanks for your hard work @alicanerdurmaz !!

### Self Check before Merge


aliemir:
I really think that we should not ship without `refine-mantine-form` or at least without a release plan for `refine-mantine-form` which wraps `@mantine/form`. React-hook-form is good to go but still `@mantine/form` will be a missing piece. 

I was going to say the same thing for the datagrid/table component but with this comment here https://github.com/mantinedev/mantine/discussions/1057#discussioncomment-3163669 So, maybe we can just use a custom table or use [`mantine-data-grid`](https://kuechlin.github.io/mantine-data-grid/) which uses react-table-v8 and at first glance looks like it's going to be compatible with out `react-table` package and combining these two will be a breeze.

PR #2543, opened 2022-09-20
Use more common expression for 24/7

rassie:
Normally, it's 24/7, not 7/24 (ref: https://en.wikipedia.org/wiki/24/7_service)


omeraplak (approved):

omeraplak:
Thank you @rassie 🚀

PR #2582, opened 2022-09-26
feat: implement auth page on refine cloud

yildirayunlu:
Support new [AuthPage](https://refine.dev/docs/api-reference/antd/components/antd-auth-page/) (login/register, etc) on refine cloud.


packages/cloud/src/hooks/useAuthProviderWithCloudConfig/index.ts
```diff
@@ -1,38 +1,132 @@
...
+                        })
+                        .catch((err) =>
+                            Promise.reject({
+                                message: "Error",
+                                name: err.message,
+                            }),
+                        );
+                },
+                forgotPassword: async ({ email }) => {
+                    return await sdk.auth
```
omeraplak, line 62: we don't need `then` and `catch` blocks

omeraplak (changes requested):

packages/sdk/src/services/auth.ts
```diff
@@ -63,13 +63,51 @@ class Auth {
...
+    async isAuthenticated(): Promise<boolean> {
+        const accessToken = this.client.getLocalStorage(
+            "refine-sdk-access-token",
+        );
+
+        if (accessToken) {
+            return true;
+        }
+
+        return false;
```
sevkioruc, line 75: Hey @yildirayunlu, could it like that as well?
```suggestion
   return !!accessToken;
```

yildirayunlu: Thank you for the suggestion @sevkioruc. It can be used. 🚀

PR #2627, opened 2022-10-04
Implement new AuthPages components on headless

yildirayunlu:

Implement new AuthPages (login, register, forgotPassword, resetPassword) components on headless.


packages/core/package.json
```diff
@@ -41,6 +41,7 @@
         "typescript": "^4.7.4"
     },
     "dependencies": {
+        "@pankod/refine-ui-types": "^0.11.2",
```
omeraplak, line 44: there seems to be something wrong here @aliemir @yildirayunlu

packages/core/src/components/pages/auth/components/updatePassword/index.tsx
```diff
@@ -1,44 +1,48 @@
...
                     e.preventDefault();
-                    if (newPassword === confirmPassword) {
-                        updatePassword({
-                            newPassword,
-                        });
-                    }
+                    updatePassword({
+                        password: newPassword,
+                        confirmPassword,
+                    });
```
omeraplak, line 43:
```suggestion
                    updatePassword({
                        password: newPassword,
                    });
```
It looks like we don't need to send the `confirmPassword`

yildirayunlu: I added it because we didn't verify that two passwords are the same in core. I think this can be check in authProvider.

omeraplak: cool 🚀

omeraplak (approved):

Issue #2631, opened 2022-10-04
[BUG] Unable to import Header component from @pankod/refine-mui


Coinio:
**Describe the bug**
When trying to import the Header component from @pankod/refine-mui into a CustomLayout I am getting the following error.

Module '"@pankod/refine-mui"' has no exported member 'Header'.ts(2305)

**To Reproduce**
1. Create a custom layout (see page https://refine.dev/docs/api-reference/mui/customization/mui-custom-layout/#creating-a-custom-layout)
2. Try to import the header component, e.g.

     `import { Header as DefaultHeader } from "@pankod/refine-mui";`

3. See error.

**Expected behavior**
Import happens without errors.

**Screenshots**
![image](https://user-images.githubusercontent.com/6752497/193830721-83d20f7a-923f-4841-abe9-ec375d74f0c9.png)

**Desktop **
 - OS: All
 - Browser N/A
 - Version ^3.45.2

Coinio:
I'll take a look at this.

omeraplak:
Hey @Coinio ,
We've released [`@pankod/refine-mui@3.48.2`](https://github.com/pankod/refine/releases/tag/%40pankod%2Frefine-mui%403.48.2). Could you approve fix?

Coinio:
@omeraplak I've just tested this and it's working fine for me now.

PR #2635, opened 2022-10-04
fix(material-ui): fix missing header export

Coinio:
This is a fix for issue #2631 which was preventing me from importing the Header component into a custom layout.


omeraplak (approved):
Thank you!

PR #2660, opened 2022-10-06
[FEAT] Add support for AND/OR conditional filters 

### LINKED ISSUE #1627: [FEAT] Add support for `and` conditional filter
Now that the conditional or filter is available (see https://github.com/pankod/refine/issues/1417) it would be cool to have a complementary `and` filter such that the following query becomes possible:

```sql
OR(
  AND(
    {title} == "The Great Divide",
    {age} < 10,
  ),
  AND(
    {title} == "The New Era",
    {age} < 20,
  )
)
```

This would allow to construct complex filters for different situations. Currently it's only possible to do the following:
```sql
AND(
  OR(
    {title} == "The Great Divide",
    {title} == "The New Era",
  ),
  OR(
    {age} < 10,
    {age} < 20,
  )
)
```

One can notice however that these are not the same and the condition for the age is now not tied to the specific titles and in this example `{age} < 10` actually being useless since `{age} < 20` already satisfies it.

An additional `and` filter based on the same principle as the `or` filter.

The top-level implicit `AND` that already exists is good but does limit the queries we can construct if we need to narrow down results individually.

Smartmind12:
Closes #1627

### DIFF
--- a/packages/core/src/contexts/data/IDataContext.ts
+++ b/packages/core/src/contexts/data/IDataContext.ts
@@ -48,7 +48,6 @@ export type CrudOperators =
     | "nbetween"
     | "null"
     | "nnull"
-    | "or"
     | "startswith"
     | "nstartswith"
     | "startswiths"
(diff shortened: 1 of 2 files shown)


alicanerdurmaz:
Hi @Smartmind12 Thanks for the PR 👊 

Unfortunately, adding only type definitions will not be enough to use this feature. We must implement `AND` logic it in the supporting dataProviders. For example: [nestjsx-crud](https://github.com/pankod/refine/blob/next/packages/nestjsx-crud/src/index.ts#L100)

Smartmind12:
> Hi @Smartmind12 Thanks for the PR 👊
> 
> Unfortunately, adding only type definitions will not be enough to use this feature. We must implement `AND` logic it in the supporting dataProviders. For example: [nestjsx-crud](https://github.com/pankod/refine/blob/next/packages/nestjsx-crud/src/index.ts#L100) is a good option to add this feature.

Hey sure will work on it and update the same!

Smartmind12:
> Hi @Smartmind12 Thanks for the PR 👊
> 
> Unfortunately, adding only type definitions will not be enough to use this feature. We must implement `AND` logic it in the supporting dataProviders. For example: [nestjsx-crud](https://github.com/pankod/refine/blob/next/packages/nestjsx-crud/src/index.ts#L100) is a good option to add this feature.

I have made the additional changes kindly review!

yildirayunlu:
Hi @Smartmind12, I made a review again. But I think it doesn't solve the problem in the issue.
You can test it with the example [here](https://github.com/pankod/refine/tree/next/examples/dataProvider/nestjsxCrud).

PR #2827, opened 2022-10-19
docs(examples-list): add ability to persist tags in url search params

aliemir:
Added search params to `docs/examples` to sync selected tags with the filter state, easy to share.

### Self Check before Merge

### DIFF
--- a/documentation/src/components/example-list/index.tsx
+++ b/documentation/src/components/example-list/index.tsx
@@ -133,20 +133,18 @@ const ExampleList: React.FC = () => {
     };
 
     const updateURLParams = (nextFilters?: Set<string>) => {
-        if (typeof history !== "undefined") {
-            if (
-                !nextFilters ||
-                nextFilters.size === visibleTags.length ||
-                nextFilters.size === 0
-            ) {
-                replace({
-                    search: "",
-                });
-            } else {
-                replace({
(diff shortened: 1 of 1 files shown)


omeraplak (approved):

documentation/src/components/example-list/index.tsx
```diff
@@ -115,6 +132,24 @@ const ExampleList: React.FC = () => {
         return PREDEFINED_COLORS[tag] || undefined;
     };
 
+    const updateURLParams = (nextFilters?: Set<string>) => {
+        if (typeof history !== "undefined") {
```
sevkioruc, line 136: Hey @aliemir, Is `"undefined"` format like that?

sevkioruc, line 136: Hey @aliemir, What is the `history`? I couldn't find its definition in this page.

aliemir: Hey @sevkioruc, it's the `window.history` (https://developer.mozilla.org/en-US/docs/Web/API/History) accessed without `window` 😅 But thanks for pointing this out, actually that check was unnecessary 🙏

sevkioruc: @aliemir Thanks for your explanation :)

PR #2838, opened 2022-10-19
fix: disabled button anchors

### LINKED ISSUE #2828: [BUG] Disable click on <*Button>s when unauthorized
### Describe the bug

Right now my "basic user" should only be able to see (not edit) a list, and its detail. 
Using `<EditButton>` in the List actions column instead makes the button appear as `disabled` (**but still clickable**).

When clicked you're (of course) redirected to a 404 page.

~~I'd also love to have a prop on the Buttons (`ListButton`, `EditButton`, `ShowButton`, `DeleteButton`) (e.g `showWhenUnauthorized` - default: `true` ) to avoid showing unauthorized buttons **at all**, as they **should not** be clicked by basic users, so it' useless to even show them~~

### Steps To Reproduce

1. Setup providers to set "can: false" on some resource/actions
2. Create a `<ShowButton>` or `<EditButton>` for given resource/action
3. Click on the button

### Expected behavior

You should

1- Not be redirected to a 404 page
~~2- (Even better) Not see the disabled button at all (maybe behind a `showWhenUnauthorized` prop)~~

### Screenshot

_No response_

### Desktop

_No response_

### Mobile

_No response_

### Additional Context

_No response_

aliemir:
Buttons with links/navigation was ignoring access provider and the disabled status.

### Closing issues

Resolves #2828

### Self Check before Merge

### DIFF
--- a/.changeset/four-bottles-kick.md
+++ b/.changeset/four-bottles-kick.md
@@ -1,0 +1,7 @@
+---
+"@pankod/refine-antd": patch
+"@pankod/refine-mantine": patch
+"@pankod/refine-mui": patch
+---
+
+Fixed #2828 - Buttons were not respecting access control when navigating to a new page. Now, if button is disabled, it will not also block the navigation not just the onClick event.
--- a/packages/antd/src/components/buttons/clone/index.tsx
+++ b/packages/antd/src/components/buttons/clone/index.tsx
@@ -78,6 +78,10 @@ export const CloneButton: React.FC<CloneButtonProps> = ({
             to={cloneUrl}
             replace={false}
             onClick={(e: React.MouseEvent<HTMLElement, MouseEvent>) => {
+                if (data?.can === false) {
+                    e.preventDefault();
+                    return;
+                }
                 if (onClick) {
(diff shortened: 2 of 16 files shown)


omeraplak (approved):

.changeset/four-bottles-kick.md
```diff
@@ -0,0 +1,7 @@
+---
+"@pankod/refine-antd": patch
+"@pankod/refine-mantine": patch
+"@pankod/refine-mui": patch
+---
+
+Fixed #2828 - Buttons were not respecting access control when navigating to a new page. Now, if button is disabled, it will not also block the navigation not just the onClick event.
```
sevkioruc, line 7: hi @aliemir, I am not sure exactly but may it be like that? :)
```suggestion
Fixed #2828 - Buttons were not respecting access control when navigating to a new page. Now, if button is disabled, it will also block the navigation not just the onClick event.
```

aliemir: @sevkioruc I think, adding _not_ will alter the meaning of the sentence here 😅

PR #2905, opened 2022-10-25
fix(icon): changed profile icon as user icon in store

### LINKED ISSUE #2777: [BUG] Updating the my profile icon
### Describe the bug

The  Profile Icon looks so simple and the icon need to something that tells the user that its the profile icon and the current icon lacks that

### Steps To Reproduce

Go to "https://store.refine.dev/"
The top uppermost right corner

### Expected behavior

Need Enhancement

### Screenshot

_No response_

### Desktop

_No response_

### Mobile

_No response_

### Additional Context

_No response_

s-vamshi:
Replaced avatar component with user icon component.

image

On hover: 
image

closes #2777 

### Self Check before Merge

### DIFF
--- a/examples/store/src/components/common/UserNav/UserNav.tsx
+++ b/examples/store/src/components/common/UserNav/UserNav.tsx
@@ -72,7 +72,7 @@ export const UserNav: React.FC<{
                                 className={s.avatarButton}
                                 onClick={() => (isSuccess ? null : openModal())}
                             >
-                                <User />
+                                <User size={24} />
                             </button>
                         </DropdownTrigger>
                         <CustomerMenuContent />
--- a/examples/store/src/components/common/UserNav/UserNav.tsx
+++ b/examples/store/src/components/common/UserNav/UserNav.tsx
@@ -6,7 +6,7 @@ import { LineItem } from "@medusajs/medusa";
 
 import { Avatar } from "@components/common";
 import { useUI } from "@lib/context";
-import { Heart, Bag, Menu } from "@components/icons";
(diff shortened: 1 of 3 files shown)


examples/store/src/components/common/UserNav/UserNav.tsx
```diff
@@ -6,7 +6,7 @@ import { LineItem } from "@medusajs/medusa";
 
 import { Avatar } from "@components/common";
```
sevkioruc, line 7: If `Avatar` don't be used other place in this file, could you please remove it?

s-vamshi: Should I remove whole file or only the import statement?

aliemir: Thanks for pointing this out @sevkioruc  🙏

examples/store/src/components/icons/User.tsx
```diff
@@ -1,14 +1,13 @@
...
-    size = "16",
     color = "currentColor",
     ...props
 }) => {
     return (
         <svg
-            width={size}
-            height={size}
+            width="22"
+            height="24"
```
aliemir, line 10: `<User/>` is also used in the profile page and this change might be an issue for it, can't we use `size` prop instead of declaring them statically? If `height=24` is for alignment maybe we can use `mt-px` or something like that?

s-vamshi: @aliemir I have added props for user icon component can you please check now?

aliemir:
Thanks for the changes @s-vamshi!

PR #2962, opened 2022-11-03
docs(blog): add esbuild post

necatiozmen:
esbuild post added


documentation/blog/2022-11-03-esbuild.md
```diff
@@ -0,0 +1,174 @@
...
+What esbuild does here is that it bundles your application into ```bundle.js```, and the whole process is extremely fast.
+
+## Is it Production ready?
+**esbuild** is a great tool with a lot of potentials, however, it is still a small project maintained by a single person. There are not a lot of open-source contributions to this project, and its author is the only person maintaining it. While **esbuild** shows great performance compared to its counterparts, being a new entrant, you will not see many projects in production with **esbuild** yet. It is better to test it on a side project and push it to production after it goes well for your need. 
+
+## esbuild TypeScript support
+For TypeScript-based projects in production, you can take advantage of using [tsup](https://github.com/egoist/tsup). 
+
+Using tsup you can build your TypeScript applications with minimal configuration.  
+It uses **esbuild** behind the scene so you get the power of esbuild along with the convenience of tsup. We, at refine, have seen remarkable performance using tsup in our projects running in production environment.
```
omeraplak, line 166:
```suggestion
It uses **esbuild** behind the scene so you get the power of esbuild along with the convenience of tsup. We, at refine, have seen remarkable performance using tsup in our project dev/build processes. 
```

PR #2983, opened 2022-11-10
feat: add generate command for refine-cli

yildirayunlu:
Add `generate` command for `refine-cli`.

### DIFF
--- a/packages/cli/src/commands/generate.ts
+++ b/packages/cli/src/commands/generate.ts
@@ -1,0 +1,19 @@
+import { Command } from "commander";
+
+const load = (program: Command) => {
+    return program
+        .command("generate:resource <name>")
+        .description("Generate a new resource")
+        .option(
+            "-a, --actions [actions]",
+            "Only generate the specified actions. (ex: list,create,edit,delete)",
+        )
+        .action(action);
+};
+
+const action = (resourceName: string, options: any) => {
+    console.log(`resourceName: ${resourceName}`);
+    console.log("options", options);
+};
+
+export default load;
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -1,6 +1,8 @@
 #!/usr/bin/env node
 import { Command } from "commander";
 
+import generate from "./commands/generate";
+
 const bootstrap = () => {
     const program = new Command();
     program
@@ -8,7 +10,8 @@ const bootstrap = () => {
         .usage("<command> [options]")
         .helpOption("-h, --help", "Output usage information.");
 
-    // TODO: load commands
+    // load commands
+    generate(program);
 
     program.parse(process.argv);
 


omeraplak (approved):

packages/cli/package.json
```diff
@@ -7,7 +7,7 @@
     "module": "dist/esm/index.js",
     "typings": "dist/index.d.ts",
     "scripts": {
-        "start": "tsup --watch --format esm,cjs,iife --legacy-output",
+        "start": "tsup --watch --format esm,cjs,iife --legacy-output && cp templates dist/templates/",
```
omeraplak, line 10: Can we do it using tsup plugin?

example: https://github.com/nickjj/esbuild-copy-static-files

yildirayunlu: I dont use this. Use from source with `../templates`.

packages/cli/src/commands/generate/resource/index.ts
```diff
@@ -0,0 +1,143 @@
...
+const load = (program: Command) => {
+    const projectType = getProjectType();
+    const { path } = getResourcePath(projectType);
+
+    return program
+        .command("generate:resource <name>")
+        .description("Generate a new resource")
+        .option(
+            "-a, --actions [actions]",
+            "Only generate the specified actions. (ex: list,create,edit,delete)",
```
omeraplak, line 22:
```suggestion
            "Only generate the specified actions. (ex: list,create,edit,show)",
```

packages/cli/src/commands/generate/resource/index.ts
```diff
@@ -0,0 +1,143 @@
...
+            defaultActions,
+        )
+        .option("-p, --path [path]", "Path to generate the resource", path)
```
omeraplak, line 25:
```suggestion
        .option("-p, --path [path]", "Path to generate the resource files", path)
```

packages/cli/src/commands/generate/resource/index.ts
```diff
@@ -0,0 +1,143 @@
...
+};
+
+const getResourcePath = (
+    projectType: ProjectTypes,
+): { path: string; alias: string } => {
+    switch (projectType) {
+        case ProjectTypes.NEXTJS:
+            return {
+                path: "src/components",
+                alias: "@components",
```
omeraplak, line 127: May cause errors in projects not created by superplate

packages/cli/src/commands/generate/resource/index.ts
```diff
@@ -0,0 +1,143 @@
...
+            return {
+                path: "app/components",
+                alias: "~/components",
+            };
+    }
+
+    // vite and react
+    return {
+        path: "src/pages",
+        alias: "pages",
```
omeraplak, line 139:
```suggestion
        alias: "pages",
```
May cause errors in projects not created by superplate

alicanerdurmaz (approved):

omeraplak (approved):

packages/cli/src/commands/generate/resource/index.ts
```diff
@@ -0,0 +1,156 @@
...
+    const jscodeshiftExecutable = require.resolve(".bin/jscodeshift");
+    const { stderr, stdout } = execa.sync(jscodeshiftExecutable, [
+        "./",
+        "--extensions=ts,tsx,js,jsx",
+        "--parser=tsx",
+        `--transform=${__dirname}/../src/transformers/resource.ts`,
+        `--ignore-pattern=**/.cache/**`,
+        `--ignore-pattern=**/node_modules/**`,
+        `--ignore-pattern=**/build/**`,
+        `--ignore-pattern=**/.next/**`,
```
aliemir, line 108: Can we read those from `.gitignore` of the target project? There might be edge cases we're missing to break the transform? 🤔 @omeraplak @yildirayunlu

yildirayunlu: Reading from `gitignore` is a good suggestion. However, in cases where this file does not exist, the cli processing times are quite long.

We can add all the cases that we support for now.

packages/cli/src/definitions/projectTypes.ts
```diff
@@ -0,0 +1,5 @@
+export enum ProjectTypes {
+    REACT_SCRIPT = "react-scripts",
+    REMIX = "remix",
+    NEXTJS = "nextjs",
+}
```
omeraplak, line 5:
```suggestion
export const projectTypes = "react-scripts" | "remix" | "nextjs"
```
should we use it like this instead of `enum`?

Issue #2624, opened 2022-10-04
[FEAT]: prepopulate initialValues in searchFormProps when syncWithLocation is true


mayank-meragi:
**Is your feature request related to a problem? Please describe.**
I am setting syncWithLocation to true for useTable from Antd, the initialValues in searchFormProps is empty

**Describe the solution you'd like**
Automatically populate initialValues in searchFormProps when syncWithLocation is true

vinayjain18:
Where this method is located in this repo @mayank-meragi ?

mayank-meragi:
@vinayjain18 https://github.com/pankod/refine/blob/next/packages/antd/src/hooks/table/useTable/useTable.ts

lex111:
@omeraplak could you please give instructions on how to resolve this issue?

omeraplak:
> @omeraplak can I work on this one?

Yes! Happy hacking 💥

toukirkhan:
@omeraplak can you provide more details to me or explain more about the issue?

omeraplak:
Hey @toukirkhan ,
The `useTable` hook allows filtering with an external form. You can view an example here. https://example.admin.refine.dev/orders?pageSize=10&current=1

When `syncWithLocation` is enabled in `useTable`, all filters start to sync with the URL, so when you share this link with someone else, the filters are applied to that client as well.

For example, when you enter this URL, pay attention to the filter form. You will see the filters in the URL come as initialValue. 
https://example.admin.refine.dev/orders?pageSize=10&current=1&filters[0][field]=q&filters[0][operator]=eq&filters[0][ value]=test&filters[1][field]=store.id&filters[1][operator]=eq&filters[1][value]=1&filters[2][field]=status.text&filters[2][operator]=in&filters[2 ][value][0]=Pending

The way to achieve this is to write the following lines, https://github.com/pankod/refine/blob/next/examples/fineFoods/admin/antd/src/pages/orders/list.tsx#L291

What is desired with this feature is that this process takes place automatically without the need to write these lines. Is everything clear to you?

toukirkhan:
@omeraplak yeah I got it!!

toukirkhan:
@omeraplak am unassigning myself as I am not able to create the output required over here. Thank you!

Mohammad-Palla:
I want to work on it. Please assign it to me

omeraplak:
> I want to work on it. Please assign it to me

Done! Happy hacking! 🚀

[...]

PR #2988, opened 2022-11-11
[FEAT]: prepopulate initialValues in searchFormProps when syncWithLocation is true

leapful:
**Explain the details for making this change. What existing problem does the pull request solve?**

When setting syncWithLocation to true for useTable from Antd, the initialValues in searchFormProps is empty.

initialValues in searchFormProps should be automatically populated when syncWithLocation is true.

**Related issues**

https://github.com/refinedev/refine/issues/2624

### Test plan (required)

1. Access Fine Foods example url at:

`https://example.admin.refine.dev/orders?pageSize=10&current=1&filters[0][field]=q&filters[0][operator]=eq&filters[0][value]=test&filters[1][field]=store.id&filters[1][operator]=eq&filters[1][value]=1&filters[2][field]=user.id&filters[2][operator]=eq&filters[2][value]=7&filters[3][field]=status.text&filters[3][operator]=in&filters[3][value][0]=Pending&filters[4][field]=createdAt&filters[4][operator]=gte&filters[4][value]=2022-11-29T17:00:00.000Z&filters[5][field]=createdAt&filters[5][operator]=lte&filters[5][value]=2022-12-27T16:59:59.999Z`

2. Check that all query parameters are filled on "Filters" menu

### Self Check before Merge


examples/fineFoods/admin/antd/src/pages/orders/list.tsx
```diff
@@ -308,7 +283,7 @@ const Filter: React.FC<{ formProps: FormProps; filters: CrudFilters }> = (
                 <Col xl={24} md={8} sm={12} xs={24}>
                     <Form.Item
                         label={t("orders.filter.status.label")}
-                        name="status"
+                        name="status.text"
```
alicanerdurmaz, line 286:
```suggestion
                        name="status"
```
This is not working because `onSearch()` still uses `status`

```
export interface IOrderFilterVariables {
    q?: string;
    store?: string;
    user?: string;
    createdAt?: [Dayjs, Dayjs];
    status?: string;
}
```

```
filters.push({
                field: "status.text",
                operator: "in",
                value: status,
});
```

alicanerdurmaz (changes requested):

examples/fineFoods/admin/antd/src/pages/orders/list.tsx
leapful: @alicanerdurmaz

Is there any reason of using field name like "store", "user", "status" instead of the real name "store.id", "user.id", "status.text" beside just alias for easy name usage ?

The purpose of this PR is to avoid mapping for initial values when using form like the previous code handling here:

Screenshot 2022-11-12 at 10 57 49

Reverting to field name "store", "user", "status" will bring back the mapping again and in terms of an example, it would be great to avoid confusion for developers by keep the default initial values (I've seen many cases that keep copy and paste the initial values mapping even that the field name is not nested field property).

In conclusion, just update "onSearch" callback is good enough:

```
onSearch: (params) => {

  filters.push({ field: 'status.text', value: params['status.text']  })
}

```

alicanerdurmaz: Hi again @leapful, Sorry for late response.
Did you test the app with these code ? It's not filtering.

Because incoming params values are these.
`{
    "status.text": [
        "Ready"
    ],
    "store.id": 2,
    "user.id": 1
}`

But you are are trying to push with value `user` not `user.id`;
```
filters.push({
                field: "user.id",
                operator: "eq",
                value: user,
            });
```

by the way, if you wan to change variable name, you need to update [interface IOrderFilterVariables](https://github.com/refinedev/refine/blob/next/examples/fineFoods/admin/antd/src/interfaces/index.d.ts) too.

alicanerdurmaz: yes this works. but you need to update the interface as well.

`onSearch: (params) => {
  filters.push({ field: 'status.text', value: params['status.text']  })
}
`

leapful: @alicanerdurmaz 

Thanks for your review. I've pushed the change and already tested successfully in local development. Please help to review it again.

examples/fineFoods/admin/antd/src/pages/orders/list.tsx
```diff
@@ -333,7 +308,7 @@ const Filter: React.FC<{ formProps: FormProps; filters: CrudFilters }> = (
                 <Col xl={24} md={8} sm={12} xs={24}>
                     <Form.Item
                         label={t("orders.filter.user.label")}
-                        name="user"
+                        name="user.id"
```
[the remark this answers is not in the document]
leapful: Resolved as it was updated after discussion.

examples/fineFoods/admin/antd/src/pages/orders/list.tsx
```diff
@@ -321,7 +296,7 @@ const Filter: React.FC<{ formProps: FormProps; filters: CrudFilters }> = (
                 <Col xl={24} md={8} sm={12} xs={24}>
                     <Form.Item
                         label={t("orders.filter.store.label")}
-                        name="store"
+                        name="store.id"
```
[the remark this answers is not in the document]
leapful: Resolved as it was updated after discussion.

alicanerdurmaz:
Hi @leapful Thank you for your great work.

We discuss with the core team and we noticed this feature can cause breaking changes.
And supporting this feature from `core` can cause too many edge cases.

Thank you again.

PR #2998, opened 2022-11-14
feat(cli): add update command

alicanerdurmaz:
Add `update` command for `refine-cli`.
refine-interactive


omeraplak (approved):

omeraplak (approved):

packages/cli/src/commands/update/interactive/index.test.tsx
```diff
@@ -0,0 +1,113 @@
+import { createUIGroup, validatePrompt } from ".";
+
+test("Validate interactive prompt", () => {
+    const testCases = [
+        {
+            input: [
```
sevkioruc, line 6: Hey @alicanerdurmaz ,
I am wondering that if these packages would been updated in the future, should been updated them in here? 
if so, maybe it need to dynamically updating in next step.

alicanerdurmaz: No, update don't needed. It's just a mock data and checking package name equality without version.
Probably I should create mock data with fake names to avaoid confusion.`ex:@owner/package-name` 
Sorry abbout that 🙏

For example;
If user choose `simple-rest` twice, we should force user to select only one package to avoid confusion.
image

sevkioruc: Hey @alicanerdurmaz,
I have just see it is a test file yet. i am soryy :)
Thanks for your clarification :)

PR #3039, opened 2022-11-22
fix TS errors when 'useParams' set to 'any

alonp99:
Please provide enough information so that others can review your pull request: quick fix for annoying ts error when using `useParams` through `routerProvider`


aliemir:
Hey @alonp99, can you add a changeset for `@pankod/refine-react-router-v6`? 

Can be added with `npm run changeset`. After that you need to select `@pankod/refine-react-router-v6`, select `patch` and add some explanation about the changes. You can check the CHANGELOG.MD files for examples. 🙏

PR #3063, opened 2022-11-25
fix: unique check on create-resource with refine-cli

yildirayunlu:
Added unique check on `create-resource` with `refine-cli`.

### DIFF
--- a/packages/cli/src/commands/create-resource/index.ts
+++ b/packages/cli/src/commands/create-resource/index.ts
@@ -5,11 +5,13 @@ import {
     moveSync,
     pathExistsSync,
     mkdirSync,
+    existsSync,
 } from "fs-extra";
 import temp from "temp";
 import { plural } from "pluralize";
 import execa from "execa";
 import inquirer from "inquirer";
+import { join } from "path";
 
 import { getProjectType, getUIFramework } from "@utils/project";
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/create-resource/index.ts
```diff
@@ -80,6 +79,21 @@ const action = async (
...
+
+        // check exist resource
+        const resourcePath = join(process.cwd(), path, resourceFolderName);
+
+        if (pathExistsSync(resourcePath)) {
+            console.error(
+                `Resource (${join(
+                    path,
+                    resourceFolderName,
+                )}) already exist! ❌`,
```
omeraplak, line 93:
```suggestion
                )}) already exist ❌`,
```

packages/cli/src/commands/create-resource/index.ts
```diff
@@ -80,6 +79,21 @@ const action = async (
 
     args.forEach((resourceName) => {
         const customActions = actions ? actions.split(",") : undefined;
+        const { path, alias } = getResourcePath(getProjectType());
+        const resourceFolderName = plural(resourceName).toLowerCase();
```
omeraplak, line 83: Can we move it to the helper function?

omeraplak (approved):

PR #3083, opened 2022-11-28
feat(cli): undefined `NODE_ENV` set to the `development` from `production`

alicanerdurmaz:

### DIFF
--- a/packages/cli/src/utils/env/index.ts
+++ b/packages/cli/src/utils/env/index.ts
@@ -1,23 +1,40 @@
 import { NODE_ENV } from "@definitions/node";
-import { env } from "node-env-type";
 import * as dotenv from "dotenv";
 dotenv.config();
 
+const envSearchMap: Record<Exclude<NODE_ENV, "custom">, RegExp> = {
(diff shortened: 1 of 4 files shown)


packages/cli/src/utils/env/index.ts
```diff
@@ -1,23 +1,40 @@
...
+
+    for (const [key, value] of Object.entries(envSearchMap)) {
+        if (value.test(nodEnv)) {
+            env = key as NODE_ENV;
+            break;
+        }
+    }
+
+    return env;
+};
```
omeraplak, line 31:
```suggestion
    const nodeEnv = process.env.NODE_ENV;

    if (!nodeEnv) {
        return "development";
    }

    let env: NODE_ENV = "custom";

    for (const [key, value] of Object.entries(envSearchMap)) {
        if (value.test(nodEnv)) {
            env = key as NODE_ENV;
            break;
        }
    }

    return env;
};
```

omeraplak: Is it a typo?

PR #3216, opened 2022-12-13
chore: support filter dropdown number value

### LINKED ISSUE #3124: [BUG] filterDropdown props under Table.column not working with single select
### Describe the bug

When I use Select without multiple mode(so single select) under filterDropdown this is not filtered.

### Steps To Reproduce

You can create single select under the filterDropDown and you will see this is not trigger re-render query.
I think reason of this is antd filter is working with array only.

### Expected behavior

When I create single select it must work properly. But nothing happening. there is no any error.

### Screenshot

_No response_

### Desktop

_No response_

### Mobile

_No response_

### Additional Context

_No response_

leapful:

By default, Ant Design expected filtered keys should be an array or string but in case value is a number, it will be treated as invalid and stop on filtering.

https://github.com/ant-design/ant-design/blob/77ea4038d169d033dcd4d29da1df9ab21b23ebb9/components/table/hooks/useFilter/FilterDropdown.tsx#L229-L233

### Test plan (required)

1. Apply filter dropdown on Select component with mode "single" and list option included number value
   For example: `{ label: "A", value: 1 }`
2. Perform select and filter on multiple items of Select component
3. Check API for filtering is called with correct selected value

Referer: https://stackblitz.com/edit/refinedev-refine-jydkgh?file=src%2Fpages%2Fposts%2Flist.tsx&preset=node

### Closing issues

closes #3124 

### Self Check before Merge

### DIFF
--- a/.changeset/silly-numbers-camp.md
+++ b/.changeset/silly-numbers-camp.md
@@ -1,0 +1,5 @@
+---
+"@pankod/refine-antd": minor
+---
+
+Support filter dropdown on number value of single Select component
--- a/packages/antd/src/components/table/components/filterDropdown/index.spec.tsx
+++ b/packages/antd/src/components/table/components/filterDropdown/index.spec.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { Input } from "antd";
-import { DatePicker } from "@components/antd";
+import { DatePicker, Select } from "@components/antd";
 import dayjs from "dayjs";
 
 import { render, fireEvent } from "@test";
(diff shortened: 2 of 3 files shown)


alicanerdurmaz:
Hi @leapful, Thans for the PR 🙏 

When <FilterDropdown/> is single mode and when we use `number` as a `value`. This code returns `array` and filter operator turns to `in`
In this example It's not a problem, but in some data providers it will be a issue and when we send filters with request body maybe we dont wanna sent number array.

Do you have any idea how to solve this problem ? When solving `Ant's` problem, it is necessary to consider all other consequences. 

Maybe <FilterDropdown> should accept `onFilter` as a prop and user's should handle this issue.

What do you think ?

leapful:
@alicanerdurmaz 

Thanks for sharing the other use cases with me :D 

When working with Ant Design Select component, I noticed that they are using string value for all handling logic even the original value is number, so may be it would be better to cast `selected keys` into string value when filtering because it's still a valid value for filter drop down from Ant Design.

For example
```
if (typeof _mappedValue === "number") {
  keys = `${_mappedValue}`;
}
```

I've tested OK where filter operator stood still "eq" instead of "in".

alicanerdurmaz:
> @alicanerdurmaz
> 
> Thanks for sharing the other use cases with me :D
> 
> When working with Ant Design Select component, I noticed that they are using string value for all handling logic even the original value is number, so may be it would be better to cast `selected keys` into string value when filtering because it's still a valid value for filter drop down from Ant Design.
> 
> For example:
> 
> ```
> if (typeof _mappedValue === "number") {
>   keys = `${_mappedValue}`;
> }
> ```
> 
> I've tested OK where filter operator stood still "eq" instead of "in".

You are right. Thank you again 🙏

omeraplak (approved):
⚡️

omeraplak:
Hey @leapful ,
Thank you! Can you add [changeset](https://refine.dev/docs/contributing/#changeset) please ?

alicanerdurmaz (approved):

PR #3224, opened 2022-12-14
chore: restore filter operator after clearing on filter dropdown

### LINKED ISSUE #1386: [BUG] Clearing filter through the clear button in filterDropdown resets operator to eq with no way to switch it back
**Describe the bug**
The clear button in the filterDropdown resets not just the value but also the operator always to "eq". This makes it impossible to use any other operator after clearing the filter.

**To Reproduce**
1. Go to a record list with filters other than `eq` (for example `contains`)
2. Filter items by something
3. Observe that the set filter operator (eg. `contains`) worked properly
4. Now click the clear button in the dropdown and then filter to get back all results
5. Observe that this works as well
6. Now open the filter again and filter by the same thing as in step 2
7. Observe that the results are now different since the filter operator switched from `contains` to `eq`

**Expected behavior**
The operator defined in `initialFilter` should not be reset when clearing a filter query. This is somewhat related to https://github.com/pankod/refine/issues/1373. Without the support for that the developer is unable to switch the operator manually so it's impossible to work around this behaviour.

**Screenshots**
refine-double-filter-contains-bug

React Query Debugging Window showing the 2 requests with the changed operator:

image

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Edge
 - Version: 2.4.11
 - Used dataProvider: Airtable

leapful:

By default, "Clear" action from Filter Dropdown is handled by Ant Design by removing filter column value out from list filter. If there is a new value on the filter column again, the mapping has no context on initial filters to restore correct operator.

By providing a map of initial filters, mapping new filters can restore operator in case previous filters is empty or no data on target filter column.

### Test plan (required)

1. Initialize example "refine-airtable-example" with initial filter on column "Title" with `{ field: 'title', operator: 'contains', value: '' }`
2. Perform filter on column "Title" with keyword "Hello" to see correct results of the keyword
3. Click "Clear" button then "Filter" button to see correct results of empty filter data
4. Filter again with keyword "Melih" to see correct results of the keyword

### Closing issues

closes #1386 

### Self Check before Merge

### DIFF
--- a/.changeset/fifty-chefs-drive.md
+++ b/.changeset/fifty-chefs-drive.md
@@ -1,5 +1,5 @@
 ---
-"@pankod/refine-core": minor
+"@pankod/refine-core": patch
 ---
 
 Restore filter operator after clear it by using filter dropdown
--- a/.changeset/fifty-chefs-drive.md
+++ b/.changeset/fifty-chefs-drive.md
@@ -1,0 +1,5 @@
+---
+"@pankod/refine-core": minor
+---
+
+Restore filter operator after clear it by using filter dropdown
--- a/packages/antd/src/hooks/table/useTable/useTable.ts
+++ b/packages/antd/src/hooks/table/useTable/useTable.ts
@@ -1,4 +1,4 @@
-import React, { Children, createElement, Fragment } from "react";
+import React, { Children, createElement, Fragment, useMemo } from "react";
 import { Grid, FormProps, Form, TablePaginationConfig, TableProps } from "antd";
 import { QueryObserverResult } from "@tanstack/react-query";
 import { useForm as useFormSF } from "sunflower-antd";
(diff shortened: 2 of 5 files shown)


IgnusG (approved):
Looks good to me 👍

packages/antd/src/hooks/table/useTable/useTable.ts
```diff
@@ -120,6 +123,15 @@ export const useTable = <
         dataProviderName,
     });
 
+    const mapInitialFilter: Record<string, CrudFilter> = useMemo(() => {
```
alicanerdurmaz, line 126: necessary to map initial filters from outside ? it's kinda hard to track.

maybe `mapAntdFilterToCrudFilter()` should accept `initialFilters` and should map inside. 

from my perspective it's better then `useMemo`

leapful: @alicanerdurmaz 

The idea is to avoid re-computing on map of initial filters but yes, it's kind hard to track in term of API function. I just updated to use directly `initialFilter` then perform mapping inside the function.

omeraplak (approved):

.changeset/fifty-chefs-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pankod/refine-core": minor
```
alicanerdurmaz, line 2:
```suggestion
"@pankod/refine-core": patch
```

alicanerdurmaz (approved):

PR #3262, opened 2022-12-19
build: update to `lerna@v6`

aliemir:
Updated `Lerna` to `v6`, we need some additional changes to enable the full potential of Lerna with Nx.


package.json
```diff
@@ -48,18 +49,17 @@
...
     "react-refresh": "^0.11.0",
     "ts-jest": "^29.0.3",
     "typescript": "^4.7.4"
   },
-  "dependencies": {
-    "@changesets/changelog-github": "^0.4.6",
-    "@changesets/cli": "^2.24.3",
-    "prettier": "^2.7.1"
+  "engines": {
+    "node": ">=10"
```
omeraplak, line 63:
```suggestion
    "node": ">=16"
```

PR #3282, opened 2022-12-21
feat(cli): npm ls should not throw error

alicanerdurmaz:
`npm -ls` always throws an error. now returns `null` and `error` handling can be done anywhere


packages/cli/src/utils/package/index.ts
```diff
@@ -92,7 +92,7 @@ export const getInstalledRefinePackagesFromNodeModules = async () => {
 
         return refinePackages;
     } catch (err) {
-        return [];
+        return Promise.resolve(null);
```
alicanerdurmaz, line 95:
```suggestion
        return [];
```

PR #3405, opened 2023-01-16
docs: add tutorial unit 6 which contains auth provider

salihozdemir:
Documentation for new tutorial. The PR is related to Unit 6 which contains Auth Provider.


alicanerdurmaz (approved):

aliemir (approved):

yildirayunlu (approved):

documentation/docs/tutorial/5-understanding-authprovider/1-create-authprovider.md
```diff
@@ -0,0 +1,953 @@
...
+<details>
+  <summary><strong>Can I pass any parameters to the <code>login</code> method?</strong></summary>
+
+Yes, you can pass any parameters to the `login` method. `useLogin` hook's mutation will pass the mutation's parameters to the `login` method without any type constraints.
+
+```ts
+const { mutate } = useLogin<{
+    username: string;
+    password: string;
+    confirmPassword: string;
```
omeraplak, line 113:
```suggestion
    foo: string;
```

documentation/docs/tutorial/5-understanding-authprovider/1-create-authprovider.md
```diff
@@ -0,0 +1,953 @@
...
+    },
+    ...
+};
+```
+
+</details>
+
+### checkAuth
+
+`checkAuth` method is used to check if the user is authenticated. Internally, it is called when the user navigates to a page that requires authentication.
```
omeraplak, line 193: and it is called page loads

documentation/docs/tutorial/5-understanding-authprovider/1-create-authprovider.md
```diff
@@ -0,0 +1,953 @@
...
+
+`checkAuth` method expects to return a Promise.
+
+-   If the Promise resolves, the user is authenticated and pages that require authentication will be accessible.
+
+-   If the Promise rejects, the user is not authenticated and pages that require authentication will not be accessible.
```
omeraplak, line 199: We should mention that when Promise.reject is returned, we redirect the user to the "/login" route and can be overridden.

documentation/docs/tutorial/5-understanding-authprovider/1-create-authprovider.md
```diff
@@ -0,0 +1,953 @@
...
+
+```tsx
+import { useAuthenticated } from "@pankod/refine-core";
+
+const { data, isSuccess, isLoading, isError, refetch } = useAuthenticated();
+```
+
+:::tip
+
+The `<Authenticated>` component makes use of the `useAuthenticated` hook. It allows you to render components only if the user is authenticated. You can see how to use the `useAuthenticated` hook in the [source code](https://github.com/refinedev/refine/blob/master/packages/core/src/components/authenticated/index.tsx).
```
omeraplak, line 235: We can avoid sharing the source code link because the file path may change in the future and we may forget to update it.

documentation/docs/tutorial/5-understanding-authprovider/1-create-authprovider.md
```diff
@@ -0,0 +1,953 @@
...
+
+For example, if you want to check if the user has a specific permission, you can use the `usePermissions` hook like this:
+
+```tsx
+import { usePermissions } from "@pankod/refine-core";
+
+const { data } = usePermissions();
+
+if (data?.includes("admin")) {
+    console.log("User has admin permission");
```
omeraplak, line 513:
```suggestion
    console.log("User has admin permissions");
```

documentation/docs/tutorial/5-understanding-authprovider/1-create-authprovider.md
```diff
@@ -0,0 +1,953 @@
...
+
+        return Promise.reject();
+    },
+    ...
+};
+```
+
+<br />
+
+Invoking the `useGetIdentity` hook will call the `getUserIdentity` method. If `getUserIdentity` method resolves a data, it will be available in the `useIdentity` hook's `data` property.
```
omeraplak, line 562: Depending on the UI framework used, we can say that when "name" and "avatar" return, we show them in the header.

documentation/docs/tutorial/5-understanding-authprovider/1-create-authprovider.md
```diff
@@ -0,0 +1,953 @@
...
+
+[Refer to the `useIdentity` documentation for more information &#8594](/docs/api-reference/core/hooks/auth/useGetIdentity/)
+
+For example, if you want to get the user's email, you can use the `useIdentity` hook like this:
+
+```tsx
+import { useIdentity } from "@pankod/refine-core";
+
+const { data } = useIdentity();
```
omeraplak, line 571:
```suggestion
import { useGetIdentity } from "@pankod/refine-core";

const { data } = useGetIdentity();
```

documentation/docs/tutorial/5-understanding-authprovider/1-create-authprovider.md
```diff
@@ -0,0 +1,953 @@
...
+}
+```
+
+### register
+
+`register` method is used to register a new user. It is similar to the `login` method. It expects to return a Promise.
+
+-   If the Promise resolves, the user is authenticated and pages that require authentication will be accessible.
+
+-   If the Promise rejects, the user is not authenticated and pages that require authentication will not be accessible.
```
omeraplak, line 584: are you sure?

documentation/docs/tutorial/5-understanding-authprovider/1-create-authprovider.md
```diff
@@ -0,0 +1,953 @@
...
+        // if request is successful, resolve the Promise, otherwise reject it
+        return Promise.resolve();
+    },
+    ...
+};
+```
+
+<br />
+
+Invoking the `useUpdatePassword` hook's mutation will call the `updatePassword` method, passing in the mutation's parameters as arguments. Additionally, the `updatePassword` method will take query parameters as arguments from the URL as well.
```
omeraplak, line 853: Can we explain with an example?

documentation/docs/tutorial/5-understanding-authprovider/headless/2-auth-pages.md
```diff
@@ -0,0 +1,426 @@
...
+                    },
+                ]}
+            />
+        );
+    };
+    ```
+
+    By default, `<AuthPage>` component renders the login page. So, we don't need to pass any props to the `<AuthPage/>` component.
+
+3. Run the app and navigate to the `/login` page.
```
omeraplak, line 143: Can we show with what parameters the login method of AuthProvider is called?

omeraplak (changes requested):

documentation/docs/tutorial/5-understanding-authprovider/1-create-authprovider.md
salihozdemir: Isn't it the same as navigating?

documentation/docs/tutorial/5-understanding-authprovider/1-create-authprovider.md
salihozdemir: Actually, `useAuthenticated` hook won't redirect to anywhere when the return Promise.reject. If user want to redirect the user specific page, I mention about it in below collapsable area.

documentation/docs/tutorial/5-understanding-authprovider/1-create-authprovider.md
salihozdemir: I think, it seems true, isn't it?

documentation/docs/tutorial/5-understanding-authprovider/1-create-authprovider.md
omeraplak: https://github.com/refinedev/refine/blob/next/packages/react-router-v6/src/routeProvider.tsx#L183

if checkAuth returns promise.reject the user is indirectly redirected to "/login"

maybe it can be mentioned like a tip or info box

documentation/docs/tutorial/5-understanding-authprovider/1-create-authprovider.md
salihozdemir: I'm not sure 😅, I'll update it.

documentation/docs/tutorial/5-understanding-authprovider/1-create-authprovider.md
omeraplak: I've always seen the use of plural

documentation/docs/tutorial/5-understanding-authprovider/1-create-authprovider.md
salihozdemir: Below, I show it with example. Can you check out it?

documentation/docs/tutorial/5-understanding-authprovider/headless/2-auth-pages.md
salihozdemir: I mentioned about it above. Can you checkout it in `Login Page` title?

omeraplak: yes but it needs to be shown more clearly

salihozdemir: I thought about it too, but we already shown an general example in first section. After we gave an example for all methods and hooks in the second section. So, I am not sure there is any need to explain again in the third part? Although we explain again with text, but we do not show code. What do you think in this context? 😅

omeraplak: AuthPage remains a bit of magic for users. We could add it to AuthPage docs. It would be nice to have a way to see what the component is calling without "debugging" it.

documentation/docs/tutorial/5-understanding-authprovider/1-create-authprovider.md
omeraplak: I'm talking about somehow explaining that you pass query parameters with an example.

yildirayunlu (commented):
No link for the next section in the `updatePassord` section.

resim

yildirayunlu (approved):

necatiozmen (approved):

omeraplak (approved):

PR #3445, opened 2023-01-23
feat: add `cursor pagination` example for mui

yildirayunlu:
Add `cursor pagination` example for `mui`


examples/cursor-pagination-mui/src/rest-data-provider/index.ts
```diff
@@ -0,0 +1,157 @@
...
+    apiUrl: string,
+    httpClient: AxiosInstance = axiosInstance,
+): Omit<
+    Required<DataProvider>,
+    "createMany" | "updateMany" | "deleteMany"
+> => ({
+    getList: async ({ resource, metaData, pagination }) => {
+        let url = `https://api.github.com/${resource}?per_page=${
+            pagination?.pageSize || 10
+        }`;
```
aliemir, line 18: Why don't we use `apiUrl` parameter? 🤔

yildirayunlu: We can use 👍🏼

aliemir (approved):

Issue #2810, opened 2022-10-18
[FEAT] Dynamic window title


omeraplak:
### Is your feature request related to a problem? Please describe.

It would be great if each page could have an auto-generated and custom title. 

Example, For the Posts resource

- dashboard -> `Dashboard | refine`
- list -> `Posts | refine`
- edit -> `#{id} Edit Post | refine`
- show -> `#{id} Show Post | refine`
- create -> `Create new Post | refine`

It can also be overridden by adding an event to the refine options.

```
onTitle: ({ resource, action, options, autoGeneratedTitle }) => {
console.log({ resource, params, options, autoGeneratedTitle });
*/
{
  resource: "posts",
  options: { foo: "bar" },
  params: { id: 1, action: "edit" }
  autoGeneratedTitle: "#{1} Edit Post | refine"
} 
*/
}
```

### Describe the thing to improve

Pages can be easily distinguished in browser history and tabs

amit-ksh:
I'd like to work on this issue. Please assign this issue to me?

omeraplak:
Done. Happy hacking 💀

amit-ksh:
@omeraplak can you give some directions like where to start to solve this issue. 🙏

omeraplak:
> @omeraplak can you give some directions like where to start to solve this issue. 🙏

Hey @amit-ksh ,
I think we can solve this problem with refine's Router Provider Packages (https://refine.dev/docs/packages/list-of-packages/).

Simply, we should update the [document title](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/title) when the user views pages with actions like `dashboard`, `list`, `create`, `edit`, and `show`.

We can create a hook (ex: `useUpdateWindowTitle`) that will update this `document title` and this hook can be exported from package `@pankod/refine-core`.

Then this hook is called in the router provider. ex: https://github.com/refinedev/refine/blob/next/packages/react-router-v6/src/routeProvider.tsx#L24

We can also add an event called `onDocumentTitle` to Refine component to enable the user to manage the title globally. This method is called before the window title change and the result returned is used (if provided).

Don't forget to review our [Contributing document](https://refine.dev/docs/contributing/)

Are these explanations enough for you?

Enjoy 🎃

amit-ksh:
So, I need to work on `react-router-v6` only and no change need to make to the `react-router` package right?

amit-ksh:
Ok 👌! Thanks, @omeraplak for the clarification

[...]

PR #3450, opened 2023-01-24
[WİP]: dynamic window title #2810

mertcanaltin:
I made a small development for the dynamic window text, I would be grateful if you have any advice on this subject. 
interacted with the refine component on an example 🚀


alicanerdurmaz:
Hello. this PR does not meet all the features of the issue

mertcanaltin:
> Hello. this PR does not meet all the features of the issue. We expect a callback where it can return a new `title` according to users' needs and it should also be managed manually with a hook. Also, It should work with all router provider packages.
> 
> If you have any further ideas or feedbacks, we can discuss on issue #2810 🙏

i'll be there in pursuit 🚀

PR #3531, opened 2023-02-03
docs(antd): update `useImport` doc

salihozdemir:
- Updated `useImport` hook doc from `@pankod/refine-antd` package.

### DIFF
--- a/documentation/docs/api-reference/antd/hooks/import/useImport.md
+++ b/documentation/docs/api-reference/antd/hooks/import/useImport.md
@@ -49,7 +49,7 @@ export const PostList: React.FC = () => {
 
 ### `resourceName`
 
-> Default: Read from the current route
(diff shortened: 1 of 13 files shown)


documentation/docs/api-reference/antd/hooks/import/useImport.md
```diff
@@ -3,158 +3,227 @@ id: useImport
...
+        </Upload>
+    );
+};
+```
+
+## Properties
+
+### `resourceName`
+
+> Default: Read from the current route
```
alicanerdurmaz, line 52:
```suggestion
> Default: It reads the `resource` value from the current URL.
```

we wrote like this on `useTable`, `useShow`, `useForm`

alicanerdurmaz (approved):

yildirayunlu (approved):

PR #3558, opened 2023-02-07
fix : typo in graphql index.ts

florent-andre:
Fixing a typo from `generete` to `generate`. As it's an exported function I don't know the impacted code. Regards.

### DIFF
--- a/packages/graphql/src/dataProvider/index.ts
+++ b/packages/graphql/src/dataProvider/index.ts
@@ -9,7 +9,7 @@ import * as gql from "gql-query-builder";
 import pluralize from "pluralize";
 import camelCase from "camelcase";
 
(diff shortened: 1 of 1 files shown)


aliemir:
Hey @florent-andre, thank you for the contribution! 🙏 

Can you check out https://refine.dev/docs/contributing/#changeset? We're using changesets for the versioning. And also, I think we can do a re-export with name `genereteExport` to avoid breaking stuff and mark it as `@deprecated`? What do you think? 🤔

aliemir:
Due to inactivity on your side, I'm closing this issue @florent-andre. Sent some updates and created an another one, you can check out and I'll be glad to hear it if you have any feedbacks/reviews 🙏 #3597

PR #3580, opened 2023-02-10
feat(v4): change resource `options`  to `meta` (#3561)

### LINKED ISSUE #3581: Additional Properties in `resource` definitions
We're not sure if we're going to keep the `resources` but if we decide to keep it in `<Refine />` component or keep it but use it in a different place, we should consider adding the ability to pass additional properties to the `resource` definitions. This will allow users to have better control in their `router` handling and other functions if we expose them in the future.

## Implementation

- Add `meta` property to `resource` interface just like `metaData` in data hooks.
- Add examples of use cases.
- Update docs to reflect the changes.

alicanerdurmaz:
`options` property of `resources` is deprecated. Added `meta` property instead. To provide backward compatibility, `options` property is still supported. But it will be removed in the next major version.

```diff
<Refine
    resources={[
        {
            name: "posts",
-           options: {},
+           meta: {},
        },
    ]}
/>
```

### Test plan (required)

All tests pass.

### Closing issues

closes #3581

### Self Check before Merge


omeraplak (approved):

aliemir (approved):

aliemir:
Let's revert the changes in `/examples` so we can test out the codemods later 🤔

aliemir (changes requested):
Let's revert the changes in `/examples` so we can test out the codemods later 🤔

aliemir (approved):

PR #3602, opened 2023-02-13
feat(v4): pagination mode

### LINKED ISSUE #3601: Removing `hasPagination` property in favor of `pagination` property
The motivation behind the `hasPagination` property was to introduce a way to disable pagination in a non-breaking way. Now with a major version, we can move this same check to `pagination` property since with `hasPagination` set to `false`, `pagination` will not be used. Instead of `hasPagination`, we can do `pagination.enabled: false` to disable pagination.

## Implementation

- Remove `hasPagination` property from `useList` hook.
- Add checks for `pagination.enabled: false` instead of `hasPagination`. (or `disabled: true`)
- Deprecate `hasPagination` property in data providers `getList` but keep it working until the next major version.
- Update docs and examples to reflect the changes.

salihozdemir:
-   `initialCurrent` and `initialPagesize` properties of `useTable` hook are deprecated. Use `pagination.current` and `pagination.pageSize` instead. However, for backward compatibility, `initialCurrent` and `initialPagesize` properties will be used if `pagination.current` and `pagination.pageSize` are not provided.

-   `sorter` and `setSorter` return values of `useTable` hook are deprecated. Use `sorters` and `setSorters` instead. However, for backward compatibility, `sorter` and `setSorter` are safe to use.

-   Added `pagination.mode` property to `useTable` hook.

    -   When `pagination.mode` is "off", all records will be fetched from the API.
    -   When `pagination.mode` is "client", all records will be fetched from the API and pagination will be handled by the `useTable` hook.
    -   When `pagination.mode` is "server", pagination will be handled by the API using `current` and `pageSize` properties of `pagination` object.

- `sorter` property of `stringifyTableParams` method is deprecated. Use `sorters` instead.

-   Now, `pagination` property of `getList` method has new `mode` property. It can be "off", "client", and "server". If it is "off" or "client", data provider will not send any pagination related query params to the API. If it is "server", data provider will send pagination related query params to the API.

-   `hasPagination` property of `getList` method is deprecated. Use `pagination.mode` instead. However, for backward compatibility, `hasPagination` property will be used if `pagination.mode` is not provided.

### Test plan (required)

All existing tests pass and added new tests.

### Closing issues

close #3601

### Self Check before Merge


packages/antd/src/hooks/list/useSimpleList/useSimpleList.ts
```diff
@@ -35,6 +36,13 @@ export type useSimpleListReturnType<
     searchFormProps: FormProps<TSearchVariables>;
     filters: CrudFilters;
     setFilters: useTableReturnType<TData>["setFilters"];
+    sorter?: CrudSorting;
+    setSorter: useTableReturnType<TData>["setSorter"];
```
omeraplak, line 40:
```suggestion
    sorters?: CrudSorting;
    setSorters: useTableReturnType<TData>["setSorter"];
```
like `filters` and `setFilters`?

omeraplak (changes requested):

aliemir (approved):

yildirayunlu (approved):

packages/airtable/src/index.ts
```diff
@@ -116,18 +116,14 @@ const AirtableDataProvider = (
...
-            //`sort` is deprecated with refine@4, refine will pass `sorters` instead, however, we still support `sort` for backward compatibility
-            const generetedSort = generateSort(sorters ?? sort) || [];
+        getList: async ({ resource, pagination, sorters, filters }) => {
+            const {
+                current = 1,
+                pageSize = 10,
+                mode = "server",
+            } = pagination ?? {};
+
+            const generetedSort = generateSort(sorters) || [];
```
BatuhanW, line 126:
```suggestion
            const generatedSort = generateSort(sorters) || [];
```

documentation/docs/api-reference/antd/hooks/table/useTable.md
```diff
@@ -52,7 +52,7 @@ const { tableProps } = useTable<IPost>();
 ```
 
 :::info
-To disable pagination, you can set `hasPagination` property to `false` which is `true` by default. If `hasPagination` has set to `false`, pagination elements will be hidden in the `<Table/>`. If you want to handle the pagination on the client-side you can override the `pagination` property in `tableProps`.
+If you want to handle the pagination on client-side, you can pass the `pagination.mode` prop to the `useTable` hook and set it to `"client"`.
```
BatuhanW, line 55: Should it be "off"?

documentation/docs/api-reference/core/hooks/useTable/index.md
```diff
@@ -30,29 +30,25 @@ In basic usage, `useTable` returns the data as it comes from the endpoint. By de
 
 ## Pagination
 
-`useTable` has a pagination feature. The pagination is done by using the `current` and `pageSize` props. The `current` is the current page and the `pageSize` is the number of records per page.
+`useTable` has a pagination feature. The pagination is done by using the `current` and `pageSize` props of `pagination`. The `current` is the current page and the `pageSize` is the number of records per page.
```
BatuhanW, line 33:
```suggestion
`useTable` has a pagination feature. The pagination is done by passing `current` and `pageSize` keys to `pagination` object. The `current` is the current page and the `pageSize` is the number of records per page.
```

documentation/docs/api-reference/core/hooks/useTable/index.md
```diff
@@ -30,29 +30,25 @@ In basic usage, `useTable` returns the data as it comes from the endpoint. By de
...
+By default, the `current` is 1 and the `pageSize` is 10. You can change default values by passing the `pagination.current` and `pagination.pageSize` props to the `useTable` hook.
 
 You can also change the `current` and `pageSize` values by using the `setCurrent` and `setPageSize` functions that are returned by the `useTable` hook. Every change will trigger a new fetch.
 
-If you want to disable pagination, you can use `hasPagination` property in `useTable` config and set it to `false`
-
-:::info
-If `hasPagination` is set to `false`, `current`, `setCurrent`, `pageSize`, `setPageSize`, and `pageCount` will return `undefined`
-:::
+If you want to handle the pagination on client-side, you can pass the `pagination.mode` prop to the `useTable` hook and set it to `"client"`.
```
BatuhanW, line 41: Should it be "off"?

documentation/docs/api-reference/antd/hooks/table/useTable.md
salihozdemir: If the user defines "off", pagination will be disabled. However, you're right because users don't handle the pagination our `useTable` hook handles the pagination. I'll update it 🤙🏻

salihozdemir: By default, pagination happens on the server side. If you want to do pagination handling on the client side, you can pass the `pagination.mode` property and set it to `"client"`. Also, you can disable the pagination by setting the `"off"`.

@BatuhanW Is it good?

.changeset/great-cooks-sell.md
```diff
@@ -0,0 +1,17 @@
...
+
+-   Added `pagination.mode` property to `useTable` and `useList` hooks.
+
+    -   When `pagination.mode` is "off", all records will be fetched from the API.
+    -   When `pagination.mode` is "client", all records will be fetched from the API and pagination will be handled by the `useTable` hook.
+    -   When `pagination.mode` is "server", pagination will be handled by the API using `current` and `pageSize` properties of `pagination` object.
+
+- `sorter` property of `stringifyTableParams` method is deprecated. Use `sorters` instead.
+
+- Added default values for `pagination.current` and `pagination.pageSize` properties of `useList` hook.
```
omeraplak, line 17: Can we share default values?

PR #3627, opened 2023-02-16
feat(v4): add codemod transformations for deprecated props

salihozdemir:
Added codemod transformation functions to fix deprecated values.


packages/codemod/src/transformations/fix-deprecated-values.ts
```diff
@@ -0,0 +1,194 @@
...
+    });
+};
+
+const resourceNametoResource = (j: JSCodeshift, source: Collection) => {
+    const resourceNames = source.find(j.Identifier, {
+        name: "resourceName",
+    });
+
+    resourceNames.forEach((path) => {
+        j(path).replaceWith(j.identifier("resource"));
```
omeraplak, line 41: Are we checking the use of resourceName in buttons?

yildirayunlu (approved):

PR #3650, opened 2023-02-23
feat(v4-core): deprecate `authProvider`, implement new one

alicanerdurmaz:
codemod debug: https://astexplorer.net/#/gist/b23540145e7209c1f952203fdec96247/latest

This is a simple change but we're thinking it will make it our `AuthProvider` interface nicer if we rename `checkAuth` to `check` and `checkError` to `onError` since its invoked when there is an error in API calls. Also might be nicer to rename `getUserIdentity` to `getIdentity` for consistency.

- Rename `checkAuth` to `check` and `checkError` to `onError` in `AuthProvider` interface.

- Rename `getUserIdentity` to `getIdentity` in `AuthProvider` interface.

- Update the internal logic to respect the changes.

- Update docs to reflect the changes.

- Update examples to reflect the changes.

-----------


packages/core/src/hooks/auth/useGetIdentity/index.ts
```diff
@@ -20,21 +56,37 @@ export type UseGetIdentityProps<TData> = {
  * @typeParam TData - Result data of the query
  *
  */
-export const useGetIdentity = <TData = any>({
+export function useGetIdentity<TData = any>({
```
BatuhanW, line 59: Why did we change this to function?

packages/core/src/contexts/auth/index.tsx
```diff
@@ -79,6 +85,98 @@ export const AuthContextProvider: React.FC<
...
+>({});
+
+export const AuthBindingsContextProvider: React.FC<
+    PropsWithChildren<IAuthBindingsContext>
+> = ({ children, isProvided, ...authBindings }) => {
+    const queryClient = useQueryClient();
+
+    const invalidateAuthStore = () => {
+        queryClient.invalidateQueries(["useAuthenticated"]);
+        queryClient.invalidateQueries(["getUserIdentity"]);
```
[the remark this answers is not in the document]
alicanerdurmaz: with new `AuthBindings`, `getUserIdentity ` renamed to `getIdentity`.
I forgot to update query key. I will 👍

packages/core/src/hooks/auth/useGetIdentity/index.ts
alicanerdurmaz: Does function overloading work with function expressions? I changed it because as far as I know it doesn't work

packages/core/src/contexts/auth/index.tsx
```diff
@@ -1,13 +1,19 @@
...
+import React, { PropsWithChildren } from "react";
 import { useQueryClient } from "@tanstack/react-query";
 
 import { useNavigation } from "@hooks";
-import { IAuthContext } from "../../interfaces";
+import { ILegacyAuthContext, IAuthBindingsContext } from "../../interfaces";
 
-export const AuthContext = React.createContext<IAuthContext>({});
+/**
+ * @deprecated `AuthContext` is deprecated with refine@4, use `AuthBindingsContext` instead, however, we still support `AuthBindingsContext` for backward compatibility.
```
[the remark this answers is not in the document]
alicanerdurmaz: typo 👍

packages/antd/src/components/layout/sider/index.tsx
```diff
@@ -99,7 +103,7 @@ export const Sider: React.FC<RefineLayoutSiderProps> = ({ render }) => {
     const logout = isExistAuthentication && (
         <Menu.Item
             key="logout"
-            onClick={() => mutateLogout()}
+            onClick={() => mutateLogout({ redirectPath: "/login" })}
```
omeraplak, line 106: This approach seems too strict. Why are we passing this in Sider?

examples/auth-antd/src/pages/dashboard.tsx
```diff
@@ -12,7 +12,9 @@ export const DashboardPage: React.FC = () => {
     }>({
         legacy: true,
     });
-    const permissions = usePermissions();
+    const permissions = usePermissions({
+        legacy: true,
```
BatuhanW, line 16: Since legacy isn't clear, let's make it something like `v3LegacyAuthProviderCompatible`, which is explicit and annoying. Encourages users to move to the new version.

BatuhanW (changes requested):

packages/antd/src/components/layout/sider/index.tsx
alicanerdurmaz: v3 `useLogout` have default `redirectPath` to `"/login"` v4 hasn't. 
I added to support the same logic for both data providers.

If we don't give default value, I think `<Sider>` users will confuse when updated to new auth provider.

packages/core/src/components/containers/refine/index.tsx
```diff
@@ -73,9 +77,15 @@ export interface RefineProps {
     dataProvider: IDataContextProvider | IDataMultipleContextProvider;
     /**
      * `authProvider` handles authentication logic like login, logout flow and checking user credentials. It is an object with methods that refine uses when necessary.
+     * @type [`AuthBindings`](/docs/api-reference/core/providers/auth-provider/)
+     */
+    authProvider?: AuthBindings;
+    /**
+     * `legacyAuthProvider` handles authentication logic like login, logout flow and checking user credentials. It is an object with methods that refine uses when necessary.
      * @type [`AuthProvider`](/docs/api-reference/core/providers/auth-provider/)
+     * @deprecated `legacyAuthProvider` is deprecated with refine@4, use `authProvider` instead, however, we still support `legacyAuthProvider` for backward compatibility.
```
salihozdemir, line 86:
```suggestion
     * @deprecated `legacyAuthProvider` is deprecated with refine@4, use `authProvider` instead.
```

packages/core/src/hooks/auditLog/useLog/index.ts
```diff
@@ -71,12 +72,15 @@ export const useLog = <
...
 
+    const authProvider = useProvidedAuthProvider();
+
     const { resources } = useContext(ResourceContext);
     const {
         data: identityData,
         refetch,
         isLoading,
     } = useGetIdentity({
+        legacy: Boolean(authProvider?.isLegacy),
```
salihozdemir, line 83: When `authProvider` is `undefined`, `legacy` will be `false`. Will this cause any problems anywhere?

alicanerdurmaz: It will be the same result as using the auth hook without `authProvider`

BatuhanW (approved):

aliemir (approved):

PR #3703, opened 2023-03-03
fix(v4): update pixels examples to v4

salihozdemir:
- Updated the "pixels" example to v4
- Updated the "pixels-admin" example to v4

### DIFF
--- a/examples/pixels-admin/src/providers/authProvider.ts
+++ b/examples/pixels-admin/src/providers/authProvider.ts
@@ -1,128 +1,225 @@
-import { LegacyAuthProvider as AuthProvider } from "@pankod/refine-core";
+import { AuthBindings } from "@pankod/refine-core";
 
 import { supabaseClient } from "utility";
 
-export const authProvider: AuthProvider = {
+export const authProvider: AuthBindings = {
     login: async ({ email, password, providerName }) => {
-        // sign in with oauth
-        if (providerName) {
-            const { data, error } = await supabaseClient.auth.signInWithOAuth({
-                provider: providerName,
+        try {
+            // sign in with oauth
+            if (providerName) {
+                const { data, error } =
+                    await supabaseClient.auth.signInWithOAuth({
+                        provider: providerName,
+                    });
+
+                if (error) {
+                    return {
+                        success: false,
+                        error,
+                    };
+                }
+
+                if (data?.url) {
+                    return {
+                        success: true,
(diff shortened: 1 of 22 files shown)


omeraplak (approved):

examples/pixels-admin/src/providers/authProvider.ts
```diff
@@ -1,128 +1,225 @@
...
+            await supabaseClient.auth.signOut();
+
+            return {
+                success: false,
+                error: new Error("User not authorized."),
+            };
+        } catch (error) {
+            return {
+                success: false,
+                error: new Error("Login failed"),
```
alicanerdurmaz, line 59:
```suggestion
                error,
```

It might be better to throw an error from catch. but I do not know. catch can send an hard to read message.

examples/pixels-admin/src/providers/authProvider.ts
```diff
@@ -1,128 +1,225 @@
...
+            }
+
+            return {
+                success: false,
+                error: new Error("Register failed"),
+            };
+        } catch (error) {
+            return {
+                success: false,
+                error: new Error("Register failed"),
```
alicanerdurmaz, line 91:
```suggestion
                error,
```

It might be better to throw an error from catch. but I do not know. catch can send an hard to read message.

examples/pixels-admin/src/providers/authProvider.ts
```diff
@@ -1,128 +1,225 @@
...
+            }
 
-        if (data) {
-            return Promise.resolve("/");
+            return {
+                success: false,
+                error: new Error("Update Password password failed"),
+            };
+        } catch (error: any) {
+            return Promise.resolve({
```
alicanerdurmaz, line 151:
```suggestion
            return ({
```

examples/pixels-admin/src/providers/authProvider.ts
```diff
@@ -1,128 +1,225 @@
...
+                redirectTo: "/login",
+            };
         }
 
-        return Promise.resolve();
+        return {
+            authenticated: true,
+        };
     },
     getPermissions: async () => {
```
alicanerdurmaz, line 198: probably need try catch

alicanerdurmaz (dismissed):

alicanerdurmaz (dismissed):

PR #3733, opened 2023-03-06
refactor(v4): codemod installs

aliemir:
Updated refine-v3-to-v4 codemod with new codemod config helper. We're now creating a temporary file and store the installations/removals there. At the final step, we're running npm commands to install and uninstall packages, then we remove the temporary file.

### Self Check before Merge


packages/codemod/src/transformations/refine3-to-refine4.ts
```diff
@@ -1,59 +1,55 @@
...
+                isOnline: true,
+            },
+        );
+    }
+
+    if (config.getUninstalls().length > 0) {
+        await remove(rootDir, config.getUninstalls(), {
+            useYarn,
+        });
+    }
```
aliemir, line 50: Spot the bug here! 🐛

BatuhanW (commented):
Screenshot 2023-03-07 at 09 30 08

BatuhanW (approved):

PR #3765, opened 2023-03-08
fix(v4): update examples starting with "w" to migrate new router provider 

salihozdemir:
Update examples starting with "u" to migrate v4.

- [x] react examples
- [x] nextjs examples
- [x] remix examples


alicanerdurmaz (dismissed):

examples/with-cypress/src/App.tsx
```diff
@@ -24,6 +24,7 @@ const App: React.FC = () => {
                         list: "/",
                         meta: {
                             icon: <DashboardOutlined />,
+                            label: "Dashboard",
```
salihozdemir, line 27: I think the `label` is not necessary, because we defined `name` property.

BatuhanW (approved):

aliemir (approved):

PR #3808, opened 2023-03-11
v4: add `useSelect` codemod

salihozdemir:
Added `useSelect` codemod to handle the `pagination.mode`.

- If `useSelect` has `hasPagination` prop, `hasPagination` will be removed instead of `pagination.mode` will be added. 
   - If `hasPagination` is true the `pagination.mode` would be "server". 
   - If `hasPagination` is false the `pagination.mode` would be "off". 
- If `useSelect` has not `hasPagination` prop, `pagination.mode` will be added.
   - `pagination.mode` would be `"server"`. Because in the previous version, the behavior is "server".


BatuhanW (approved):
🔥 🔥 🔥

PR #3907, opened 2023-03-17
fix(mui): contrast text color for white theme was white

### LINKED ISSUE #3897: [BUG] contrastText on lightTheme same as darkTheme
### Describe the bug

I've noticed this when a user with white theme (oh boy) accessed my ReadyPage (swizzled from the default one).
The LightTheme values for `contrastText` are white - when I think they're suppose to be black (or some shade-of).
https://github.com/refinedev/refine/blob/next/packages/mui/src/theme/palette/lightPalette.ts

### Steps To Reproduce

Well you can probably see it in the default ReadyPage with White Theme

### Expected behavior

Should be shade of black.

### Screenshot

_No response_

### Desktop

_No response_

### Mobile

_No response_

### Additional Context

_No response_

YechiamTK:

### Test plan (required)

### Closing issues
closes #3897

### Self Check before Merge

### DIFF
--- a/packages/mui/src/theme/palette/lightPalette.ts
+++ b/packages/mui/src/theme/palette/lightPalette.ts
@@ -4,11 +4,11 @@ export const lightPalette: PaletteOptions = {
     mode: "light",
     primary: {
         main: "#67be23",
-        contrastText: "#fff",
+        contrastText: "#000",
     },
     secondary: {
         main: "#2A132E",
-        contrastText: "#fff",
+        contrastText: "#000",
     },
     background: {
         default: "#f0f0f0",
@@ -16,19 +16,19 @@ export const lightPalette: PaletteOptions = {
     },
     success: {
         main: "#67be23",
-        contrastText: "#fff",
+        contrastText: "#000",
     },
     error: {
         main: "#fa541c",
-        contrastText: "#fff",
+        contrastText: "#000",
     },
     warning: {
         main: "#fa8c16",
-        contrastText: "#fff",
+        contrastText: "#000",
     },
     info: {
         main: "#0b82f0",
-        contrastText: "#fff",
+        contrastText: "#000",
     },
     divider: "rgba(0,0,0,0)",
     text: {


packages/mui/src/theme/palette/lightPalette.ts
```diff
@@ -4,31 +4,31 @@ export const lightPalette: PaletteOptions = {
     mode: "light",
     primary: {
         main: "#67be23",
-        contrastText: "#fff",
+        contrastText: "#000",
```
alicanerdurmaz, line 7: Hi @YechiamTK, thanks for the PR.

Our example app text colors broken with this settings.

image

because purple and black do not have contrast.
```ts
  primary: {
        main: "#67be23",
        contrastText: "#000",
    },
```

Can you control these changes again? You can use this tool for debugging https://webaim.org/resources/contrastchecker/

alicanerdurmaz (changes requested):

packages/mui/src/theme/palette/lightPalette.ts
YechiamTK: So, looking again more thoroughly at the light theme colors; I see that primary, secondary, success and warning colors are all the same for dark and light. Is this on purpose? Shouldn't there be a different shade of these colors for the light theme?

alicanerdurmaz: the purpose of contrast text is to make sure that the text is distinguishable from its background and legible to the reader.
some UI libraries make this automatically according to the given color.

I created an example with our CSS file. I didn't see a problem.
https://codepen.io/AlicanErdurmaz/pen/LYJXbEb

What exactly was the problem you encountered? Can you share the use case?

YechiamTK: @alicanerdurmaz ok I see, I did not understand correctly the meaning of contrastText; my bad. You can reject this PR and I'll close the issue :)

PR #3911, opened 2023-03-17
fix: `useModalForm`'s `autoSubmitClose` and `autoResetForm` features

salihozdemir:
`autoSubmitClose` and `autoResetForm` were not working even though they had these properties.

Now, both of them work properly.

### DIFF
--- a/.changeset/smooth-planets-suffer.md
+++ b/.changeset/smooth-planets-suffer.md
@@ -2,5 +2,5 @@
 "@refinedev/core": patch
 ---
 
-Now, when the form is submitted successfully, the `setId` function is not called.
+In forms that use `useForm`, the `onFinish` will reset the current `id` to `undefined` when the mutation is successful. Now, the `id` will not be set to `undefined`.
 
--- a/.changeset/smooth-planets-suffer.md
+++ b/.changeset/smooth-planets-suffer.md
@@ -1,0 +1,6 @@
+---
+"@refinedev/core": patch
+---
+
+Now, when the form is submitted successfully, the `setId` function is not called.
+
--- a/.changeset/smooth-planets-suffer.md
+++ b/.changeset/smooth-planets-suffer.md
@@ -2,5 +2,5 @@
 "@refinedev/core": patch
 ---
 
-In forms that use `useForm`, the `onFinish` will reset the current `id` to `undefined` when the mutation is successful. Now, the `id` will not be set to `undefined`.
+In forms that use `useForm`, the `onFinish` was resetting the current `id` to `undefined` when the mutation is successful. Now, the `id` will not be set to `undefined`.
 
--- a/packages/antd/src/hooks/form/useModalForm/index.spec.tsx
+++ b/packages/antd/src/hooks/form/useModalForm/index.spec.tsx
@@ -1,6 +1,6 @@
 import { renderHook, waitFor } from "@testing-library/react";
 
-import { act, TestWrapper } from "@test";
+import { act, MockJSONServer, TestWrapper } from "@test";
 
 import { useModalForm } from "./";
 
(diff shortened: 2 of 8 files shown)


BatuhanW (changes requested):
Let's add some tests

packages/antd/src/hooks/form/useModalForm/index.spec.tsx
```diff
@@ -0,0 +1,181 @@
...
+            {
+                wrapper: TestWrapper({}),
+            },
+        );
+
+        await act(async () => {
+            result.current.modalProps.onCancel?.({} as any);
+        });
+
+        expect(result.current.modalProps.open).toBe(false);
```
BatuhanW, line 53:
```suggestion
        expect(result.current.modalProps.open).toBe(true);

        await act(async () => {
            result.current.modalProps.onCancel?.({} as any);
        });

        expect(result.current.modalProps.open).toBe(false);
```

packages/antd/src/hooks/form/useModalForm/index.spec.tsx
```diff
@@ -0,0 +1,181 @@
...
+            {
+                wrapper: TestWrapper({}),
+            },
+        );
+
+        await act(async () => {
+            result.current.show();
+        });
+
+        expect(result.current.modalProps.open).toBe(true);
```
BatuhanW, line 72:
```suggestion
        expect(result.current.modalProps.open).toBe(false);

        await act(async () => {
            result.current.show();
        });

        expect(result.current.modalProps.open).toBe(true);
```

packages/antd/src/hooks/form/useModalForm/useModalForm.ts
```diff
@@ -247,6 +208,9 @@ export const useModalForm = <
         sunflowerUseModal.show();
     }, []);
 
+    const { visible: _visible, ...otherModalProps } = modalProps;
```
omeraplak, line 211: Is this a breaking change? Why did we change this?

packages/antd/src/hooks/form/useModalForm/index.spec.tsx
```diff
@@ -0,0 +1,181 @@
...
+                foo: "bar",
+            });
+            result.current.modalProps.onOk?.({} as any);
+        });
+
+        await waitFor(() => result.current.mutationResult.isSuccess);
+
+        expect(result.current.formProps.form?.getFieldsValue()).toStrictEqual(
+            {},
+        );
```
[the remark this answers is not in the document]
salihozdemir: In this hook, `isDirty` doesn't affect the working logic of the `useModalForm` hook so we should not add it.

.changeset/smooth-planets-suffer.md
```diff
@@ -0,0 +1,6 @@
+---
+"@refinedev/core": patch
+---
+
+Now, when the form is submitted successfully, the `setId` function is not called.
```
BatuhanW, line 5: I think this is an internal knowledge and won't give an idea about who is reading it. Perhaps we can mention the fixed behaviour? That would be better, if someone else has this problem, it will be clear that it's fixed.

omeraplak (approved):

alicanerdurmaz (approved):

.changeset/smooth-planets-suffer.md
```diff
@@ -0,0 +1,6 @@
+---
+"@refinedev/core": patch
+---
+
+In forms that use `useForm`, the `onFinish` will reset the current `id` to `undefined` when the mutation is successful. Now, the `id` will not be set to `undefined`.
```
BatuhanW, line 5:
```suggestion
In forms that use `useForm`, the `onFinish` was resetting the current `id` to `undefined` when the mutation is successful. Now, the `id` will not be set to `undefined`.
```

BatuhanW (approved):

PR #3941, opened 2023-03-21
fix(react-table): fixed bug that `current` being 1 when the `syncWithLocation` enabled

salihozdemir:
Fixed an issue where `current` was 1 when `filters` and `sorters` were in the URL when `syncWithLocation` was enabled.

### DIFF
--- a/packages/react-table/src/useTable/index.ts
+++ b/packages/react-table/src/useTable/index.ts
@@ -1,4 +1,4 @@
-import { useEffect, useRef } from "react";
+import { useEffect } from "react";
 import {
     BaseRecord,
     CrudOperators,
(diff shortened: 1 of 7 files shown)


packages/react-table/src/useTable/index.ts
```diff
@@ -176,3 +178,13 @@ export function useTable<
         refineCore: useTableResult,
     };
 }
+
+function useFirstRender() {
```
omeraplak, line 182: Can we take this to a separate file? Also, let's not forget to add the tests.

omeraplak (changes requested):

packages/react-table/src/utils/useFirstRender/index.ts
```diff
@@ -0,0 +1,11 @@
+import { useEffect, useRef } from "react";
+
+export const useFirstRender = () => {
```
alicanerdurmaz, line 3: what do you think ? 🤔 

```suggestion
export const useIsFirstRender = () => {
```

alicanerdurmaz (approved):

omeraplak (changes requested):

omeraplak (approved):

PR #3954, opened 2023-03-23
fix(mantine): reset dirty state after setting initial values

salihozdemir:
Fixed an issue where the dirty state was not reset after setting initial values. This caused the form to be dirty even though changes were not made. For this reason, the UnSavedChangesNotifier was always shown when the user tried to leave the page.

### DIFF
--- a/.changeset/chatty-squids-approve.md
+++ b/.changeset/chatty-squids-approve.md
@@ -2,4 +2,4 @@
 "@refinedev/mantine": patch
 ---
 
-Fixed an issue where the form dirty state was not reset after setting initial values. This caused the form to be dirty even though changes were not made. For this reason, the `<UnSavedChangesNotifier>` was always warn when the user tried to leave the page.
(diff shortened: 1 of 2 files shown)


.changeset/chatty-squids-approve.md
```diff
@@ -0,0 +1,5 @@
+---
+"@refinedev/mantine": patch
+---
+
+Fixed an issue where the form dirty state was not reset after setting initial values. This caused the form to be dirty even though changes were not made. For this reason, the `<UnSavedChangesNotifier>` was always warn when the user tried to leave the page.
```
BatuhanW, line 5:
```suggestion
Fixed an issue where the form dirty state was not reset after setting initial values. This caused the form to be dirty even though changes were not made. For this reason, the `<UnSavedChangesNotifier>` always warned when user tried to leave page.
```

BatuhanW (approved):

PR #3972, opened 2023-03-24
Add missing 's' to filters

rikas:
Very basic typo fix.

I tried to use filters here (Refine 4) and it needs to be "filters" not "filter".


omeraplak (approved):
Thank you! 💥

salihozdemir (approved):

PR #4009, opened 2023-03-28
fix(live-previews): missing icons in live previews

aliemir:
Added missing icons from `@tabler/icons` and `@ant-design/icons`.

Only adding the ones that are used in `create-refine-app` templates.

### Self Check before Merge


aliemir:
Checked the docs for the fallback icon usages and replaced them with the actual imports.

PR #4017, opened 2023-03-29
feat(live-previews): `react-dom` and `web-vitals` mocks

aliemir:
To be able to run the apps from `create-refine-app`, we need to provide mock implementations for `react-dom` and `web-vitals` which are not used in live previews but present in the generated code.

Also improved the dynamic import block a bit. 

And replaced the loader.

### Self Check before Merge


packages/live-previews/pages/preview.tsx
```diff
@@ -23,109 +23,120 @@ const Preview: NextPage = () => {
...
+                        ? (await import("../src/scope/antd")).default
+                        : {},
+                async () =>
+                    usedPackages.has("mui")
+                        ? (await import("../src/scope/mui")).default
+                        : {},
+                async () =>
+                    usedPackages.has("mantine")
+                        ? (await import("../src/scope/mantine")).default
+                        : {},
```
BatuhanW, line 38: I think we have have an array of objects and extract these lines from here like:

```
const packagesMap = [{name: "mui", path: "../src/scope/mui" }]
```

BatuhanW (approved):

PR #4071, opened 2023-04-03
fix: update `authBindings` error type

salihozdemir:
- [x] Core changes
- [x] Docs
- [x] Examples
- [x] Superplate


documentation/docs/packages/documentation/routers/remix-router.md
```diff
@@ -1066,7 +1072,7 @@ Default paths are:
 -   `create`: `/resources/create`
 -   `edit`: `/resources/edit/:id`
 -   `show`: `/resources/show/:id`
-:::
+    :::
```
omeraplak, line 1075:
```suggestion
:::
```

omeraplak (changes requested):

.changeset/metal-worms-explode.md
```diff
@@ -0,0 +1,5 @@
+---
+"@refinedev/core": patch
+---
+
+Update `authBindings` error type to provide changeable error messages on notifcations.
```
omeraplak, line 5: Can we add an example of the usage here?

packages/core/src/interfaces/bindings/auth.ts
```diff
@@ -31,23 +31,25 @@
  * Same goes for `onError` function, it should always resolve.
  */
 
+import { HttpError } from "../HttpError";
```
omeraplak, line 34: Perhaps instead of using HttpError here, we can create a type called RefineError and use it. We can extend HttpError from RefineError. HttpError can be used in DataProvider and RefineError can be used in places like this.

omeraplak (changes requested):

omeraplak (approved):

PR #4122, opened 2023-04-07
fix(mantine): unexpected alert on `warnWhenUnsavedChanges`

alicanerdurmaz:
-   Fixed: `warnWhenUnsavedChanges` was throwing even when there were no changes.


examples/table-mantine-advanced/package.json
```diff
@@ -16,6 +16,7 @@
         "@refinedev/react-table": "^5.1.4",
         "@tanstack/react-table": "^8.2.6",
         "@mantine/rte": "^5.4.1",
```
salihozdemir, line 18:
```suggestion
```

You need to remove its dependency on all examples 🙃

salihozdemir (approved):

examples/table-mantine-advanced/package.json
alicanerdurmaz: DONE

PR #4123, opened 2023-04-07
refactor: updated the generic type name of hooks that use `useQuery`

salihozdemir:
Updated the generic type name of hooks that use `useQuery` to synchronize generic type names with `tanstack-query`.

`TData` => `TQueryFnData`
`TSelectData` => `TData`

### DIFF
--- a/.changeset/wet-otters-bake.md
+++ b/.changeset/wet-otters-bake.md
@@ -1,0 +1,9 @@
+---
+"@refinedev/antd": patch
+"@refinedev/core": patch
+"@refinedev/mantine": patch
+"@refinedev/mui": patch
+"@refinedev/react-table": patch
+---
+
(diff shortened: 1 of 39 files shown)


.changeset/wet-otters-bake.md
```diff
@@ -0,0 +1,9 @@
+---
+"@refinedev/antd": patch
+"@refinedev/core": patch
+"@refinedev/mantine": patch
+"@refinedev/mui": patch
+"@refinedev/react-table": patch
+---
+
+Updated the generic type name of hooks that use `useQuery` to synchronize generic type names with `tanstack-query`.
```
omeraplak, line 9: Could we add usage of example?

salihozdemir: The usage of generics didin't change. Just I updated the generic names.

omeraplak (approved):

PR #4132, opened 2023-04-10
feat: add `ThemedLayoutContext` and `useSiderVisible` hook

yildirayunlu:
`<Sider>` visibility has been improved with support for `ThemedLayoutContext` and `useSiderVisible` hook.


packages/core/src/contexts/themedLayout/index.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { ReactNode, useState } from "react";
+
+import { IThemedLayoutContext } from "./IThemedLayoutContext";
+
+export const ThemedLayoutContext = React.createContext<IThemedLayoutContext>(
+    {},
+);
+
+export const ThemedLayoutContextProvider: React.FC<{ children: ReactNode }> = ({
```
omeraplak, line 9: I think we should define this Context in UI frameworks. @BatuhanW @aliemir @yildirayunlu

yildirayunlu: It is common for all of them. Why?

omeraplak: The core package should be distributed independently of the UI. This way, we can avoid including unnecessary components for setups that do not use Sider (headless)

PR #4148, opened 2023-04-12
fix(docs): added config provider to antd edit page

DoguhanOzgurAkca:
fix(docs): `configProvider` was missing from the antd edit page example, so i added it. 
Screenshot 2023-04-12 at 12 03 07

Screenshot 2023-04-12 at 12 02 55


alicanerdurmaz (approved):

DoguhanOzgurAkca:
> Do we have any other broken example?

I am currently going through the docs so i will probably see if there is any
So far, i looked at the chakra ui, headless and the rest of the antd ones and all seem to be working.

BatuhanW (changes requested):
Let's also add missing import to example codeblock below.

BatuhanW (approved):

PR #4159, opened 2023-04-12
fix(core): missing resource `meta` in route creation

aliemir:
When creating the routes with parameters internally, we were missing the `resource.meta` values which causes missing parameters when user defines a default parameter value in `resource.meta` and doesn't have any values in current `meta` or `params`.

### Self Check before Merge

### DIFF
--- a/packages/core/src/definitions/helpers/router/compose-route.ts
+++ b/packages/core/src/definitions/helpers/router/compose-route.ts
@@ -1,5 +1,4 @@
-import { MetaDataQuery, ParseResponse } from "../../../interfaces";
-
+import { MetaQuery, ParseResponse } from "src/interfaces";
 import { pickRouteParams } from "./pick-route-params";
 import { prepareRouteParams } from "./prepare-route-params";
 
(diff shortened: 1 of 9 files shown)


alicanerdurmaz (approved):

packages/core/src/hooks/navigation/index.ts
```diff
@@ -76,7 +84,15 @@ export const useNavigation = () => {
             }
 
             return go({
-                to: composeRoute(createActionRoute, parsed, meta),
+                to: composeRoute(
+                    createActionRoute,
+                    {},
```
BatuhanW, line 89: Why don't we pass parsed now?

packages/core/src/hooks/router/use-get-to-path/index.ts
```diff
@@ -49,7 +49,15 @@ export const useGetToPath = (): GetToPathFn => {
                 return undefined;
             }
 
-            const composed = composeRoute(actionRoute, parsed, meta);
+            const composed = composeRoute(
+                actionRoute,
+                {},
```
BatuhanW, line 54: Seems like we always pass empty object here -at least in this PR-, should we remove this param if it's not needed?

BatuhanW:
Would be nice to have some tests.

packages/core/src/hooks/router/use-get-to-path/index.ts
aliemir: Updated with a change in `composeRoute` helper.

packages/core/src/definitions/helpers/router/compose-route.ts
```diff
@@ -12,13 +11,25 @@ import { prepareRouteParams } from "./prepare-route-params";
...
     // prepareRouteParams (from route params, params and meta)
-    const preparedRouteParams = prepareRouteParams(routeParams, params, meta);
+    const preparedRouteParams = prepareRouteParams(routeParams, {
+        ...resourceMeta,
+        ...(typeof parsed?.id !== "undefined" ? { id: parsed?.id } : {}),
+        ...(typeof parsed?.action !== "undefined"
+            ? { action: parsed?.action }
+            : {}),
+        ...(typeof parsed?.resource !== "undefined"
+            ? { resource: parsed?.resource }
```
BatuhanW, line 28:
```suggestion
            ? { resource: parsed.resource }
```

BatuhanW (approved):

PR #4185, opened 2023-04-17
feat(docs): added markdown feature to tutorial tracker

DoguhanOzgurAkca:
Made it possible to have markdown on the tutorial tracker 
from:
Screenshot 2023-04-17 at 14 48 23
to:
Screenshot 2023-04-17 at 14 43 09

It works for *cursive*, **bold**, ***cursive+bold*** and `code`. Additional markdowns can be added to the  `markdownConverter` function if needed
For the main titles(Create Data Provider with `swizzle` in this case) to work and convert properly, they need to be brought down from the front matter and placed below it with a # below. I am planning to go through all titles in tutorial soon anyways, so i can do those changes myself when i do that. 
If there is a way to keep them in the front matter yet use markdown, that can be implemented too but the Docusaurus discord said that it is not possible  
Screenshot 2023-04-17 at 14 49 15

### DIFF
--- a/documentation/docs/tutorial/2-understanding-dataprovider/1-swizzle.md
+++ b/documentation/docs/tutorial/2-understanding-dataprovider/1-swizzle.md
@@ -1,12 +1,13 @@
 ---
 id: swizzle
-title: 2. Create Data Provider with Swizzle
 tutorial:
     prev: tutorial/understanding-dataprovider/index
     next: tutorial/understanding-dataprovider/create-dataprovider
 ---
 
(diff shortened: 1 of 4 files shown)


DoguhanOzgurAkca:
I think it is possible to do it with refine markdown as well but it seems like it would overcomplicate things, so i just used `dangerouslySetInnerHTML`.

aliemir:
Updated the markdown converter with a proper parser

omeraplak (approved):

PR #4188, opened 2023-04-17
feat: add `ThemedLayoutv2` for `@refinedev/mantine`

yildirayunlu:
Add `ThemedLayoutv2` for `@refinedev/mantine`


aliemir:
Old `ThemedLayout` is not listed in the swizzle config. @yildirayunlu

aliemir (approved):

yildirayunlu:
> 

Yes, it was never added. I did not add this PR because it is deprecated. 
We recommend using ThemedLayoutV2 🚀

PR #4211, opened 2023-04-20
feat: allow custom input types for graphql data provider

### LINKED ISSUE #3409: [FEAT] Graphql - configurable mutation InputType

In `create` and `update` mutation queries, `operation` name can be configured via `metadata`. `variable.input.type` can't be configured as it's "hard coded". 

There is case (or at list mine ;)), where `InputType.type` can't be the same as the "hardcoded by refine" ones.

_No response_

### Additional context

_No response_

For example, in the [dataProvider.create](https://github.com/refinedev/refine/blob/master/packages/graphql/src/dataProvider/index.ts#L125), and in the same spirit that allow to configure `operation` : 
```
const operation = metaData?.operation ?? camelCreateName;
```
The variables.input.type should be something like: 
```
const variableInputType = metaData?.variableInputType ?? `${camelCreateName}Input`
```

What do you think about that ? Is it doable ? 
Thanks for this great piece of lib !

mattbho:
The custom GraphQL data provider hard codes input types. 

This PR adds a way for the user to put their own type via the `meta` field when instantiating the data provider.

Example:

```
const { data } = await dataProvider(client).create({
    resource: "posts",
    variables: {
        title: "foo",
        content: "bar",
        category: "2",
    },
    meta: {
        inputType: <input type here>
        fields: [
            {
                operation: "post",
                fields: [
                    "id",
                    "title",
                    "content",
                    { category: ["id"] },
                ],
                variables: {},
            },
        ],
    },
});

```

### Test plan (required)

I didn't include tests initially because this is an optional parameter for a custom GraphQL database. That means the field could really be anything. Of course, I am open to any requested changes and will add any needed coverage.

### Closing issues

closes #3409

### Self Check before Merge

### DIFF
--- a/packages/graphql/src/dataProvider/index.ts
+++ b/packages/graphql/src/dataProvider/index.ts
@@ -73,13 +73,14 @@ const dataProvider = (client: GraphQLClient): Required<DataProvider> => {
             const camelCreateName = camelCase(`create-${singularResource}`);
 
             const operation = meta?.operation ?? camelCreateName;
+            const inputType = meta?.inputType ?? `${camelCreateName}Input`;
 
             const { query, variables: gqlVariables } = gql.mutation({
                 operation,
                 variables: {
                     input: {
                         value: { data: variables },
-                        type: `${camelCreateName}Input`,
+                        type: inputType,
                     },
                 },
                 fields: meta?.fields ?? [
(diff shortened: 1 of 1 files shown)


BatuhanW (approved):
Hey @mattbho, this looks great as always! 🚀 

Optionally, would you like to add some documentation and/or example about this, so users will know about it?

PR #4213, opened 2023-04-21
fix(supabase): return data on create/update

### LINKED ISSUE #4212: [BUG] Supabase not receiving back data on insert/update
### Describe the bug

Supabase (PostgREST) requires explicitly asking for data back when creating or updating using `.select()`, which is not currently present.

### Steps To Reproduce

1. Setup a project with supabase as data provider
2. Use the `useCreate` (or `useUpdate`) hook
3. The data, both the awaited value and the `onSuccess` callback, are undefined

### Expected behavior

The data is returned as inserted (or updated).

### Screenshot

_No response_

### Desktop

_No response_

### Mobile

_No response_

### Additional Context

Adding `.select()` (https://supabase.com/docs/reference/javascript/insert) is required to set the `Prefer: return=representation` header that PostgREST requires to actually return data (https://postgrest.org/en/stable/api.html#insertions).

02JanDal:

Getting data back when inserting or updating using supabase requires an extra function call, see #4212 for details.

See #4212 for details.

### Test plan (required)

Tests previously had invalid mocks. They have been fixed accordingly.

### Closing issues

Closes #4212

### Self Check before Merge


aliemir (commented):
Hey @02JanDal thank you for the PR! Are there any specific reason for `select` to be different in `create` and `update`? It looks like `create` and `createMany` has a default `"*"` in the selection set but `update` and `updateMany`  only appends `select` if its defined 🤔 I like the approach in `update` methods more actually 😅 

Also any chance to write some tests for `update` and `updateMany` as well when you have the time?

aliemir (changes requested):
Let's decide on which approach we should follow and update the methods accordingly. And have simple tests for update method as well

omeraplak (approved):
LGTM! 🎯

aliemir (approved):

alicanerdurmaz (approved):

PR #4240, opened 2023-04-27
docs: `isSticky` docs for all ui libs

yildirayunlu:
`isSticky` docs for all ui libs

### DIFF
--- a/documentation/docs/api-reference/antd/components/themed-layout.md
+++ b/documentation/docs/api-reference/antd/components/themed-layout.md
@@ -257,7 +257,7 @@ const App: React.FC = () => {
 };
 ```
 
-You can also make it sticky with the `isSticky` component. It is optional and is `false` by default. The usage example is below.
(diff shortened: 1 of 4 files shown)


documentation/docs/api-reference/antd/components/themed-layout.md
```diff
@@ -257,6 +257,35 @@ const App: React.FC = () => {
 };
 ```
 
+You can also make it sticky with the `isSticky` component. It is optional and is `false` by default. The usage example is below.
```
omeraplak, line 260:
```suggestion
You can also make it `sticky` with the `isSticky` property. It is optional and is `false` by default. The usage example is below.
```

omeraplak (changes requested):

alicanerdurmaz (approved):

aliemir (approved):

omeraplak (approved):

PR #4265, opened 2023-05-03
fix: fix an issue that caused the duplicate field error

salihozdemir:
Fixed an issue that caused the duplicate field error.


aliemir (approved):

aliemir:
Did we complete the tests with inferencer examples, especially the graphql one? @salihozdemir

PR #4267, opened 2023-05-04
fix: `useModalForm` and `useDrawerForm` use `onFinish` prop

yildirayunlu:
Form submit problem when `onFinish` is overridden in `useModalForm` and `useDrawerForm`

### DIFF
--- a/documentation/docs/api-reference/antd/hooks/form/useDrawerForm.md
+++ b/documentation/docs/api-reference/antd/hooks/form/useDrawerForm.md
@@ -372,39 +372,44 @@ It renders `<Drawer>` instead of lazy rendering it.
 
 You may need to modify the form data before it is sent to the API.
 
-For example, Let's send the values we received from the user in two separate inputs, name and surname, to the API as fullName.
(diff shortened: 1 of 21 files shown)


alicanerdurmaz (approved):

documentation/docs/api-reference/antd/hooks/form/useDrawerForm.md
```diff
@@ -366,6 +366,52 @@ Current visible state of `<Drawer>`.
...
+export const UserCreate: React.FC = () => {
+    // highlight-start
+    const { formProps, drawerProps, saveButtonProps, onFinish } = useDrawerForm({
+        action: "create",
+    });
+    // highlight-end
+
+    // highlight-start
+    const handleOnFinish = (values) => {
+        onFinish({
```
salihozdemir, line 391:
```suggestion
        formProps.onFinish({
```

It was working with `formProps.onFinish`, are you sure it works with `onFinish` directly? Because in the hook you only changed the `onFinish` method in `formProps`.

salihozdemir (approved):

PR #4283, opened 2023-05-08
docs: fix wrong hash for options property

Mr0nline:
Currently it's redirecting to `#options-1` which is not accessible because `options` should points to `#options`.

### DIFF
--- a/packages/core/src/components/containers/refine/index.tsx
+++ b/packages/core/src/components/containers/refine/index.tsx
@@ -173,7 +173,7 @@ export interface RefineProps {
     onLiveEvent?: LiveModeProps["onLiveEvent"];
     /**
      * `options` is used to configure the app.
-     * @type [`IRefineOptions`](/docs/api-reference/core/components/refine-config/#options-1)
+     * @type [`IRefineOptions`](/docs/api-reference/core/components/refine-config/#options)
      * */
     options?: IRefineOptions;
 }


omeraplak (approved):
LGTM! 🎯 Thank you

packages/core/src/components/containers/refine/index.tsx
```diff
@@ -173,7 +173,7 @@ export interface RefineProps {
     onLiveEvent?: LiveModeProps["onLiveEvent"];
     /**
      * `options` is used to configure the app.
-     * @type [`IRefineOptions`](/docs/api-reference/core/components/refine-config/#options-1)
+     * @type [`IRefineOptions`](/docs/api-reference/core/components/refine-config/#options)
```
omeraplak, line 176:
```suggestion
     * @type [`IRefineOptions`](https://refine.dev/docs/api-reference/core/components/refine-config/#options)
```

omeraplak (changes requested):

Mr0nline:
This is my first open source contribution so I didn't know which branch to choose so I've just followed github's flow. Let me know if any changes require from my end! :raised_hands:

packages/core/src/components/containers/refine/index.tsx
```diff
@@ -173,7 +173,7 @@ export interface RefineProps {
     onLiveEvent?: LiveModeProps["onLiveEvent"];
     /**
      * `options` is used to configure the app.
-     * @type [`IRefineOptions`](/docs/api-reference/core/components/refine-config/#options-1)
+     * @type [`IRefineOptions`](https://refine.dev/docs/api-reference/core/components/refine-config/#options)
```
Mr0nline, line 176: I've applied the changes to use `https://refine.dev`.

Shouldn't we apply the same for every properties above as well?

omeraplak: Yes, you're right, it should be done throughout the entire file :)

Mr0nline: On it :)

Mr0nline: @omeraplak done :+1:

omeraplak (approved):

PR #4298, opened 2023-05-08
fix(live-previews): add max-http-header to start script

BatuhanW:
added `--max-http-header-size` to live previews start and dev scripts to prevent 431 status code.


omeraplak (approved):

Realworld112:
How do I learn

PR #4306, opened 2023-05-09
feat: `syncWithLocation.syncId` default to `true` for `useDrawerForm` and `useModalForm`

yildirayunlu:
`syncWithLocation.syncId` default to `true` for `useDrawerForm` and `useModalForm`

### DIFF
--- a/.changeset/curvy-carpets-push.md
+++ b/.changeset/curvy-carpets-push.md
@@ -1,5 +1,5 @@
 ---
-"@refinedev/mantine": patch
+"@refinedev/mantine": minor
 ---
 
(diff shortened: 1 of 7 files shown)


omeraplak (approved):

.changeset/curvy-carpets-push.md
```diff
@@ -0,0 +1,5 @@
+---
+"@refinedev/mantine": patch
```
aliemir, line 2: If we're marking this as `feat`, I think we should at least mark the version bump as `minor`. @omeraplak

PR #4313, opened 2023-05-10
feat: dynamic window title 

abdellah711:
- add `generateDocumentTitle` option to Refine component; when enabled, Refine will generate titles in the same manner described in this issue #2810 
```ts
options={{
    warnWhenUnsavedChanges: true,
    syncWithLocation: true,
    generateDocumentTitle: true
}}
```
> Note: the default value is `false`
- `generateDocumentTitle` also accepts a callback function that will get invoked whenever the url changes, and it should return a string that will be used to set the window title
```ts
options = {{
    warnWhenUnsavedChanges: true,
        syncWithLocation: true,
            generateDocumentTitle: ({
                resource, // posts
                action, // create
                params, // {id: 1}
                autoGeneratedTitle // Create new Post | Refine
            }) => {
                return `Refine ${action}` // => 'Refine create' 
            }
}}
```
- add `useDocumentTitle` hook which allows the user to change the title from any component
```ts
// to update the title as soon as the component is rendered
useDocumentTitle('New Post')

// to update the title at a later point use the updateTitle function returned from the hook
const {
    title, // current document title
    updateTitle
    } = useDocumentTitle()

updateTitle("New Post")

``

### CODE REVIEW

**Review by aliemir — CHANGES_REQUESTED** [review]
Hey @abdellah711, fantastic work! Thank you for the contribution. 

We have couple of changes in mind, let me try to explain each of them about why and how. We're sorry because we didn't updated the task description in time 🤦 

We're keeping the `@refinedev/core` available for every platform that can run `react` (e.g: `react-native`). The `RouteChangeHandler` and similar components are legacy from the v3 and older. Instead of placing this inside the core package, we can move your components and logic inside the router packages with small changes. 

### `DocumentTitleHandler`

Let's move the document title change logic into router packages with a component like `DocumentTitleHandler` (any suggestions on name are welcome). We can place this component inside `<Refine>`'s children to let it use the hooks and features of refine. 

It can accept the same type in props as the one you've defined in the `<Refine>` components `options` prop. 

**Example**

```tsx L5
const App = () => {
  return (
    <BrowserRouter>
      <Refine>
        <Routes>
          /* ... */
        </Routes>
        <DocumentTitleHandler handler={({ ... }) => "My Title"} />
      </Refine>
    </BrowserRouter>
  )
}
```

`handler` is optional and if you don't provide one then we can use the auto generated one. 

The example above uses the `@refinedev/react-router-v6` as an example but this will also be available in `@refinedev/nextjs-router` and `@refinedev/remix-router`. 

### `useDocumentTitle` 

This hook looks useful but can be implemented with an optional context. Not sure if the context is necessary for persisting the title value since the `<title>` won't be re-rendered. We should also try to implement an unmount logic to restore the title (if not overwritten by anywhere else). This might be tricky to implement, we can have a discussion on how to do it.

### `i18n`

`<Refine>` accepts `i18nProvider` and uses it when placing texts if possible. [Here](https://github.com/refinedev/refine/blob/0ffe70308b24d2d70695399fb0a1b7b76bcf2ccb/packages/antd/src/components/buttons/create/index.tsx#L100)'s an example in `@refinedev/antd`'s `CreateButton`, using `useTranslate` hook to render the label. 

We need to have this functionality in this feature as well. We can use a key scope like `documentTitle` and do stuff like: `documentTitle.default`, `documentTitle.show` and `documentTitle.suffix` etc. 

This will let our users just modify auto generated titles as they like without having to create a custom logic for it. 

### Changesets

We use changesets to keep the logs for our versions. Can you please provide them for the packages you've worked on. You can check out the [docs/contributing#changeset](https://refine.dev/docs/contributing/#changeset) on how to add one.

### Documentation 

We'd love to mention this feature on our docs. It will be great if you can add a section for it in the router docs. 

### Changes in `@refinedev/react-router-v6` package

I see that you've changed the router bindings but the changes you've done looks like they are only providing the params when there are `resource` and `action` defined. This won't always be the case when developing apps with refine. Users can create pages without explicitly assigning them to a `resource` and an `action` in `<Refine>`, then pass `resource` and `action` values explicitly in hooks and components as they like. 

With the changes requested above, this changes might not be required but just wanted to let you know 😅 

----

Let us know if you have any problems updating your implementation according to requirements above. 🙏 

Keep up the good work 🚀

abdellah711:
Hi @aliemir, thank you for your valuable feedback. I appreciate your point, and I will make the necessary changes as soon as possible. However, I'd like to mention two things:

- You mentioned that the context is not necessary for persisting the title and I agree with you. However, I only utilized the context to prioritize title updates from the `useDocumentTitle` hook over the auto-generated ones. This is because, based on my understanding, the useEffect of child components executes before the parent's, which mean it will overwrite the title set by the child component.

- Regarding the change I made to the `@refinedev/react-router-v6` package: I found that `useParams` doesn't function outside of a `Route` component. That's why I replaced it with `matchPath`, but this necessitated checking if the `resource` and `action` (in other words, the params that I want to match ) were defined before executing the function.So I guess I should only use `matchPath` only if `useParams` has returned `undefined` ( which means it's been called outside of `Route` ). If you have any alternative solutions, please let me know.

Thank you once again for your feedback, and I'll be waiting for your opinion on the points I mentioned.

aliemir:
> Hi @aliemir, thank you for your valuable feedback. I appreciate your point, and I will make the necessary changes as soon as possible. However, I'd like to mention two things:
> 
> * You mentioned that the context is not necessary for persisting the title and I agree with you. However, I only utilized the context to prioritize title updates from the `useDocumentTitle` hook over the auto-generated ones. This is because, based on my understanding, the useEffect of child components executes before the parent's, which mean it will overwrite the title set by the child component.
> * Regarding the change I made to the `@refinedev/react-router-v6` package: I found that `useParams` doesn't function outside of a `Route` component. That's why I replaced it with `matchPath`, but this necessitated checking if the `resource` and `action` (in other words, the params that I want to match ) were defined before executing the function.So I guess I should only use `matchPath` only if `useParams` has returned `undefined` ( which means it's been called outside of `Route` ). If you have any alternative solutions, please let me know.
> 
> Thank you once again for your feedback, and I'll be waiting for your opinion on the points I mentioned.

You're right, `useDocumentTitle` when used in multiple ways in the component tree may cause inconsistencies/mismatches. I actually meant to say that, instead of using the context in the core package, we can export it in our router packages and let the user wrap the `<Refine>` component if they want to. I thought maybe if the user wants to provide dynamic titles using `useDocumentTitle` and also set auto generated titles to `false`. Which in a way might make the context unnecessary. Does it make sense? We should still export the context though 😅 

About the change in `@refinedev/react-router-v6`, yes it makes sense to use `matchPath` as a fallback and let users use `useParams` without `resource` and `action` if they want to. I think `useParams` doesn't throw an error, when used outside of `<Route>`, right? 🤔

aliemir:
I'm looking forward to see your updates on this PR 👀 I'm excited about how this PR will turn out. 🚀

abdellah711:
Hi @aliemir, I've made the changes you requested. Here's a summary of the updates:

### `DocumentTitleHandler`
The `DocumentTitlteHandler` is now exported from `react-router-v6` and `nextjs-router`. However, I couldn't create one for `remix` and `appdir` in NextJS since their approach to handling document titles is outside the component itself.
### `useDocumentTitle`
I've used `useLayoutEffect` in the `DocumentTitleHandler` to ensure that it updates the title before `useDocumentTitle`. This way, the hook can override the title set by the component.
### `i18n`
I've introduced the key scope `documentTitle`. This scope includes the following keys:
```json
    "documentTitle": {
        "default": "Refine",
        "suffix": " | Refine",
        "post": {
            "list": "Posts | Refine",
            "show": "#{{id}} Show Post | Refine",
            "edit": "#{{id}} Edit Post | Refine",
            "create": "Create new Post | Refine",
        }
    }
```
`default`: is used when the current route is not a `resource`
`suffix`: Appended to the default generated titles.
`[resource]`:  An object containing titles for each action within the resource.

The `useDocumentTitle` hook also supports i18n by accepting an object instead of a string, with an i18nKey property. For example:
```ts
useDocumentTitle({i18nKey: "documentTitle.default"})
```

### Changeset
I tried to use changeset as mentioned in the docs but I got the following error:
```
> changeset

🦋  error Error: Failed to find where HEAD diverged from master. Does master exist?
```

### Documentation
Once you confirm that these changes are as desired, I would be glad to update the documentation.

Thank you!

abdellah711:
I think the problem came when I forked the repository; I accidentally chose to copy only the `next` branch. Any idea how I can fetch the `master` branch from the original repository? 
Thanks

abdellah711:
I tried `git fetch --all`, but unfortunately it didn't work because the `master` branch doesn't exist in the remote repository (the forked one ).

aliemir:
> I tried `git fetch --all`, but unfortunately it didn't work because the `master` branch doesn't exist in the remote repository (the forked one ).

I think you can do try adding the main repo as a remote and checkout/push master 🤔 

```
git remote add refineOrigin https://github.com/refinedev/refine.git
git fetch refineOrigin
git checkout master
git push origin master
```

This way we'll make sure we have `master` in `origin` (your fork). Then, you can checkout to `feat/dynamic-window-title` and changesets should work as expected 🙏

abdellah711:
> > I tried `git fetch --all`, but unfortunately it didn't work because the `master` branch doesn't exist in the remote repository (the forked one ).
> 
> I think you can do try adding the main repo as a remote and checkout/push master 🤔
> 
> ```
> git remote add refineOrigin https://github.com/refinedev/refine.git
> git fetch refineOrigin
> git checkout master
> git push origin master
> ```
> 
> This way we'll make sure we have `master` in `origin` (your fork). Then, you can checkout to `feat/dynamic-window-title` and changesets should work as expected 🙏

That did work! Thank you.

Please let me know if there are any further modifications that need to be done.

aliemir:
Hey @abdellah711, thank you for the changes 🙏 

Can you add the necessary section to the docs? 

- https://refine.dev/docs/packages/documentation/routers/react-router-v6/#additional-components here for React Router v6 
- https://refine.dev/docs/packages/documentation/routers/nextjs/#additional-components and here for Next.js

Also if you have time, can you create separate changesets for core and router packages? One for each package so we can give more details about the changes 🚀

abdellah711:
Hi @aliemir, I've made the changes you asked me for, and I also merged the `next` branch into the `feat/dynamic-window-title` branch.

Please let me know if I did anything wrong or if you need any additional modifications.

Thank you!

BatuhanW (approved):
Hello @abdellah711, this looks great! My friend @aliemir will do a final round of testing locally just in case we have some edge cases.

Great work 🚀

documentation/docs/api-reference/core/providers/i18n-provider.md
```diff
@@ -313,6 +313,16 @@ Before we get started, let's look at which parts are going to be translated:
...
+    },
+    "documentTitle": {
+        "default": "Refine",
+        "suffix": " | Refine",
+        "post": {
+            "list": "Posts | Refine",
+            "show": "#{{id}} Show Post | Refine",
+            "edit": "#{{id}} Edit Post | Refine",
+            "create": "Create new Post | Refine",
+        }
```
omeraplak, line 325:
```suggestion
        "default": "refine",
        "suffix": " | refine",
        "post": {
            "list": "Posts | refine",
            "show": "#{{id}} Show Post | refine",
            "edit": "#{{id}} Edit Post | refine",
            "create": "Create new Post | refine",
        }
```

documentation/docs/api-reference/core/providers/i18n-provider.md
```diff
@@ -476,6 +486,16 @@ values={[{ label: "English", value: "en" }, { label: "German", value: "de" }]}>
...
```
omeraplak, line 498:
```suggestion
    "documentTitle": {
        "default": "refine",
        "suffix": " | refine",
        "post": {
            "list": "Posts | refine",
            "show": "#{{id}} Show Post | refine",
            "edit": "#{{id}} Edit Post | refine",
            "create": "Create new Post | refine",
        }
```

documentation/docs/api-reference/core/providers/i18n-provider.md
```diff
@@ -623,6 +643,16 @@ values={[{ label: "English", value: "en" }, { label: "German", value: "de" }]}>
...
+    },
+    "documentTitle": {
+        "default": "Refine",
+        "suffix": " | Refine",
+        "post": {
+            "list": "Beiträge | Refine",
+            "show": "#{{id}} Beitrag anzeigen | Refine",
+            "edit": "#{{id}} Beitrag bearbeiten | Refine",
+            "create": "Neuen Beitrag erstellen | Refine"
+        }
```
omeraplak, line 655:
```suggestion
    "documentTitle": {
        "default": "refine",
        "suffix": " | refine",
        "post": {
            "list": "Beiträge | refine",
            "show": "#{{id}} Beitrag anzeigen | refine",
            "edit": "#{{id}} Beitrag bearbeiten | refine",
            "create": "Neuen Beitrag erstellen | refine"
        }
```

documentation/docs/packages/documentation/routers/nextjs-router.md
```diff
@@ -546,6 +546,127 @@ This feature is not working in experimental `appDir` mode in Next.js due to limi
...
+:::
+
+This component will generate the document title for the current page.By default, it follows a set of predefined rules to generate titles based on the provided props. However, it also offers the flexibility to customize the title generation process by providing a custom `handler` function.
+The default title generation rules are as follows:
+
+-   list    : `Posts | Refine`
+-   edit    : `#{id} Edit Post | Refine`
+-   show    : `#{id} Show Post | Refine`
+-   create  : `Create new Post | Refine`
+-   default : `Refine`
```
omeraplak, line 563:
```suggestion
-   list    : `Posts | refine`
-   edit    : `#{id} Edit Post | refine`
-   show    : `#{id} Show Post | refine`
-   create  : `Create new Post | refine`
-   default : `refine`
```

documentation/docs/packages/documentation/routers/nextjs-router.md
```diff
@@ -546,6 +546,127 @@ This feature is not working in experimental `appDir` mode in Next.js due to limi
...
+This hook also returns a function that can be used to set the document title dynamically.
+
+```tsx
+import { useDocumentTitle } from "@refinedev/nextjs-router";
+
+const PostList = () => {
+    const setTitle = useDocumentTitle();
+
+    useEffect(() => {
+        setTitle("Posts | Refine");
```
omeraplak, line 659:
```suggestion
        setTitle("Posts | refine");
```

documentation/docs/packages/documentation/routers/react-router-v6.md
```diff
@@ -689,6 +689,117 @@ const App = () => {
...
+### `DocumentTitleHandler`
+
+This component will generate the document title for the current page.By default, it follows a set of predefined rules to generate titles based on the provided props. However, it also offers the flexibility to customize the title generation process by providing a custom `handler` function.
+The default title generation rules are as follows:
+
+-   list    : `Posts | Refine`
+-   edit    : `#{id} Edit Post | Refine`
+-   show    : `#{id} Show Post | Refine`
+-   create  : `Create new Post | Refine`
+-   default : `Refine`
```
omeraplak, line 701:
```suggestion
-   list    : `Posts | refine`
-   edit    : `#{id} Edit Post | refine`
-   show    : `#{id} Show Post | refine`
-   create  : `Create new Post | refine`
-   default : `refine`
```

documentation/docs/packages/documentation/routers/react-router-v6.md
```diff
@@ -689,6 +689,117 @@ const App = () => {
...
+
+### `useDocumentTitle`
+This hook allows you to set the document title for the current page. It can be used in any component that is a child of the `<Refine>` component.
+
+```tsx
+import { useDocumentTitle } from "@refinedev/react-router-v6";
+
+const PostList = () => {
+    // highlight-next-line
+    useDocumentTitle("Posts | Refine");
```
omeraplak, line 757:
```suggestion
    useDocumentTitle("Posts | refine");
```

documentation/docs/packages/documentation/routers/react-router-v6.md
```diff
@@ -689,6 +689,117 @@ const App = () => {
...
+This hook also returns a function that can be used to set the document title dynamically.
+
+```tsx
+import { useDocumentTitle } from "@refinedev/react-router-v6";
+
+const PostList = () => {
+    const setTitle = useDocumentTitle();
+
+    useEffect(() => {
+        setTitle("Posts | Refine");
```
omeraplak, line 792:
```suggestion
        setTitle("Posts | refine");
```

packages/core/src/definitions/helpers/generateDocumentTitle/index.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    });
+
+    it("should return `resource name` when action is `list`", () => {
+        expect(
+            generateDefaultDocumentTitle(
+                translateMock,
+                { name: "posts" },
+                "list",
+            ),
+        ).toBe("Posts | Refine");
```
omeraplak, line 21:
```suggestion
        ).toBe("Posts | refine");
```

packages/core/src/definitions/helpers/generateDocumentTitle/index.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    });
+
+    it("should return `Create new resource name` when action is `create`", () => {
+        expect(
+            generateDefaultDocumentTitle(
+                translateMock,
+                { name: "posts" },
+                "create",
+            ),
+        ).toBe("Create new Post | Refine");
```
omeraplak, line 31:
```suggestion
        ).toBe("Create new Post | refine");
```

packages/core/src/definitions/helpers/generateDocumentTitle/index.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    it("should return `#id Edit resource name` when action is `edit`", () => {
+        expect(
+            generateDefaultDocumentTitle(
+                translateMock,
+                { name: "posts" },
+                "edit",
+                "1",
+            ),
+        ).toBe("#1 Edit Post | Refine");
```
omeraplak, line 42:
```suggestion
        ).toBe("#1 Edit Post | refine");
```

packages/core/src/definitions/helpers/generateDocumentTitle/index.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    it("should return `#id Show resource name` when action is `show`", () => {
+        expect(
+            generateDefaultDocumentTitle(
+                translateMock,
+                { name: "posts" },
+                "show",
+                "1",
+            ),
+        ).toBe("#1 Show Post | Refine");
```
omeraplak, line 53:
```suggestion
        ).toBe("#1 Show Post | refine");
```

packages/core/src/definitions/helpers/generateDocumentTitle/index.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+        generateDefaultDocumentTitle(
+            translateMock,
+            { name: "posts" },
+            "show",
+            "1",
+        ),
+            expect(translateMock).toHaveBeenCalledWith(
+                "documentTitle.posts.show",
+                { id: "1" },
+                "#1 Show Post | Refine",
```
omeraplak, line 66:
```suggestion
                "#1 Show Post | refine",
```

packages/core/src/definitions/helpers/generateDocumentTitle/index.ts
```diff
@@ -0,0 +1,48 @@
...
+        list: "",
+    };
+
+    const resourceName = userFriendlyResourceName(
+        resource?.name,
+        action === "list" ? "plural" : "singular",
+    );
+
+    const defaultTitle = translate("documentTitle.default", "Refine");
+    const suffix = translate("documentTitle.suffix", " | Refine");
```
omeraplak, line 34:
```suggestion
    const defaultTitle = translate("documentTitle.default", "refine");
    const suffix = translate("documentTitle.suffix", " | refine");
```

omeraplak (changes requested):
Could we update "Refine" to "refine"?

documentation/docs/packages/documentation/routers/nextjs-router.md
```diff
@@ -546,6 +546,127 @@ This feature is not working in experimental `appDir` mode in Next.js due to limi
...
+:::
+
+This hook allows you to set the document title for the current page. It can be used in any component that is a child of the `<Refine>` component.
+
+```tsx
+import { useDocumentTitle } from "@refinedev/nextjs-router";
+
+const PostList = () => {
+    // highlight-next-line
+    useDocumentTitle("Posts | Refine");
```
aliemir, line 624:
```suggestion
    useDocumentTitle("Posts | refine");
```

.changeset/flat-rockets-call.md
```diff
@@ -0,0 +1,13 @@
...
+
+feat: implement `generateDefaultDocumentTitle` function
+
+This function generates a default document title based on the current route by following these rules (`resource` in this case is "Post"):
+
+-   list -> `Posts | Refine`
+-   edit -> `#{id} Edit Post | Refine`
+-   show -> `#{id} Show Post | Refine`
+-   create -> `Create new Post | Refine`
+-   default (not a `resource`) -> `Refine`
```
aliemir, line 13:
```suggestion
-   list -> `Posts | refine`
-   edit -> `#{id} Edit Post | refine`
-   show -> `#{id} Show Post | refine`
-   create -> `Create new Post | refine`
-   default (not a `resource`) -> `refine`
```

.changeset/thick-trees-repair.md
```diff
@@ -0,0 +1,26 @@
...
+
+The `DocumentTitleHandler` component accepts an optional prop called `handler`, which is a callback function. This function is triggered whenever the `pathname` changes and receives an object with the following parameters:
+
+```ts
+{
+    resource, // 'posts'
+    action, // 'create'
+    params, // {id: 1}
+    pathname, // '/posts/create'
+    autoGeneratedTitle; // 'Create new Post | Refine'
```
aliemir, line 17:
```suggestion
    autoGeneratedTitle; // 'Create new Post | refine'
```

aliemir (approved):

omeraplak (approved):

aliemir:
Thank you for your effort @abdellah711 🚀

abdellah711:
I'm glad I could help, and thank you as well for allowing me to be part of your project. I look forward to working together on new features. 🚀

PR #4388, opened 2023-05-24
test(examples): add e2e test to tanstack table examples and handsome-table example 

salihozdemir:
Added e2e tests to tanstack-table examples.

### Self Check before Merge

### DIFF
--- a/examples/table-chakra-ui-basic/cypress/e2e/all.cy.ts
+++ b/examples/table-chakra-ui-basic/cypress/e2e/all.cy.ts
@@ -1,0 +1,154 @@
+/// <reference types="cypress" />
+/// <reference types="../../cypress/support" />
+
+describe("table-chakra-ui-basic", () => {
+    beforeEach(() => {
+        cy.interceptGETPosts();
+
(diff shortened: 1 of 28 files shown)


examples/table-chakra-ui-basic/cypress/e2e/all.cy.ts
```diff
@@ -0,0 +1,154 @@
...
+        cy.get(".icon-tabler-filter").eq(0).click();
+        cy.get("#title").type("lorem");
+        cy.get(".icon-tabler-check").click();
+
+        cy.url().should(
+            "include",
+            `filters[0][field]=title&filters[0][operator]=contains&filters[0][value]=lorem`,
+        );
+
+        cy.wait("@getFilteredPosts");
```
BatuhanW, line 99: Can we assert all listed rows has lorem text?

salihozdemir: We can't, I'm just checking if the parameter goes correctly in the request, so it doesn't matter.

PR #4404, opened 2023-05-26
refactor: fix name and state inconsistency in `<ThemedLayoutV2>`.

salihozdemir:
`useSiderVisible` is deprecated, instead we created a new hook `useSiderState` for it. `useSiderState` similar to `useSiderVisible` but it returns more meaningful state names. However, `useSiderVisible` is still available for backward compatibility.

Updated `Sider` and `HamburgerMenu` components using `useSiderState`.

```tsx
import { useSiderState } from '@refinedev/antd'; // or @refinedev/chakra-ui, @refinedev/mui, @refinedev/mantine
const {
    siderCollapsed,
    setSiderCollapsed,
    mobileSiderOpen,
    setMobileSiderOpen,
} = useSiderState();
```

### Test plan (required)

Tests not affected.

### Self Check before Merge


packages/chakra-ui/src/hooks/useSiderState/index.ts
```diff
@@ -0,0 +1,22 @@
+import { useContext } from "react";
+
+import { ThemedLayoutContext } from "../../contexts/themedLayoutContext";
+import { IThemedLayoutContext } from "../../contexts/themedLayoutContext/IThemedLayoutContext";
+
+export type UseSiderStateType = IThemedLayoutContext;
+
+export const useSiderState = (): UseSiderStateType => {
```
alicanerdurmaz, line 8: maybe we can change `useSiderState` name to `useSider` or `useThemedSider`.
with that name convention, we can add more features to this hook in the future.

@aliemir @BatuhanW @yildirayunlu what do you think?

alicanerdurmaz (approved):

packages/chakra-ui/src/hooks/useSiderState/index.ts
omeraplak: it's a good idea

salihozdemir: `useSiderState` has more predictable than `useSider` or `useThemedSider`. I mean, the users can better know what the `useSiderState` hook does.

If we want to change the name to add something in the future we can do something more generic like `useLayoutContext`.

aliemir: Agree with @salihozdemir, `useSiderState` is more descriptive than `useSider` or `useThemedSider` which looks more like `useMenu` or `useBreadcrumb` and users might confuse them with `useMenu`. 

We can call it `useThemedLayoutContext`, which matches with the name of the context as well (`ThemedLayoutContext`).

aliemir:
### General

- [ ] We've resolved the inconsistency issues throughout the UI packages
- [ ] State values are renamed properly and usable through the new hook
- [ ] Old values are deprecated but still functioning
- [ ] Implemented the `useThemedLayoutContext` hook.

### `@refinedev/antd`

- [ ] Context is updated and uses the new names for the states.
- [ ] Old values are deprecated but aligned with the new state values.
- [ ] Users with swizzled siders can keep using without breaking changes + no need to update the variables if they don't want to.
- [ ] Users with swizzled layout can keep using the old context provider component without breaking changes or any need for an update. This means the `initialSiderCollapsed` still works as expected.
- [ ] Users can use new values and adjust their components to use them with no issues. 
- [ ] We've updated the examples with the new hook and the new values. 
- [ ] The old hooks is still working but deprecated.

### `@refinedev/mui`

- [ ] Context is updated and uses the new names for the states.
- [ ] Old values are deprecated but aligned with the new state values.
- [ ] Users with swizzled siders can keep using without breaking changes + no need to update the variables if they don't want to.
- [ ] Users with swizzled layout can keep using the old context provider component without breaking changes or any need for an update. This means the `initialSiderCollapsed` still works as expected.
- [ ] Users can use new values and adjust their components to use them with no issues. 
- [ ] We've updated the examples with the new hook and the new values. 
- [ ] The old hooks is still working but deprecated.

### `@refinedev/chakra-ui`

- [ ] Context is updated and uses the new names for the states.
- [ ] Old values are deprecated but aligned with the new state values.
- [ ] Users with swizzled siders can keep using without breaking changes + no need to update the variables if they don't want to.
- [ ] Users with swizzled layout can keep using the old context provider component without breaking changes or any need for an update. This means the `initialSiderCollapsed` still works as expected.
- [ ] Users can use new values and adjust their components to use them with no issues. 
- [ ] We've updated the examples with the new hook and the new values. 
- [ ] The old hooks is still working but deprecated.

### `@refinedev/mantine`

- [ ] Context is updated and uses the new names for the states.
- [ ] Old values are deprecated but aligned with the new state values.
- [ ] Users with swizzled siders can keep using without breaking changes + no need to update the variables if they don't want to.
- [ ] Users with swizzled layout can keep using the old context provider component without breaking changes or any need for an update. This means the `initialSiderCollapsed` still works as expected.
- [ ] Users can use new values and adjust their components to use them with no issues. 
- [ ] We've updated the examples with the new hook and the new values. 
- [ ] The old hooks is still working but deprecated.

----

@salihozdemir Let's use the above checklist to make sure we're good to go.

aliemir (approved):

aliemir (approved):

PR #4426, opened 2023-05-31
test: increate core/helpers test coverage

yildirayunlu:
Increased test coverage of core/helpers

### Test plan (required)

Screenshot 2023-05-31 at 15 21 52

### Self Check before Merge


packages/core/src/definitions/table/index.spec.ts
```diff
@@ -602,4 +617,23 @@ describe("definitions/table", () => {
             ),
         ).toBe(true);
     });
+
+    it("parseTableParams default sorter and filters", () => {
+        expect(parseTableParams(`?current=1&pageSize=10`)).toStrictEqual({
```
salihozdemir, line 622: We can add filters and sorters to check them

yildirayunlu: Actually, I tested the case defined here as the default when `filters` and `sorters` are not given.

Screenshot 2023-06-01 at 09 41 41

alicanerdurmaz (approved):

PR #4430, opened 2023-06-01
feat(core): add `queryMeta` and `mutationMeta` properties to `useForm`

aliemir:
Added `queryMeta` and `mutationMeta` properties to the `useForm` hook. These properties are used to pass specific meta values to the query or mutation. This is useful when you have overlapping values in your data provider's `getOne` and `update` methods. For example, you may want to change the `method` of the mutation to `PATCH` but if you pass it in the `meta` property, you'll end up changing the method of the `getOne` request as well.

`queryMeta` and `mutationMeta` has precedence over `meta`. This means that if you have the same property in `queryMeta` and `meta`, the value in `queryMeta` will be used.

**Usage**

```tsx
import { useForm } from "@refinedev/core";

export const MyEditPage = () => {
    const form = useForm({
        // this is passed both to the mutation and the query requests
        meta: {
            myValue: "myValue"
        },
        // this is only passed to the query request
        queryMeta: {
            propertyOnlyWorksForQuery: "propertyOnlyWorksForQuery"
        },
        // this is only passed to the mutation request
        mutationMeta: {
            propertyOnlyWorksForMutation: "propertyOnlyWorksForMutation"
        }
    })
}
```

### Test plan (required)

Added test cases to check for properly sending `queryMeta`, `mutationMeta` and their precedence over `meta` property.

### Self Check before Merge


.changeset/forty-nails-remember.md
```diff
@@ -0,0 +1,5 @@
+---
+"@refinedev/core": patch
```
yildirayunlu, line 2: We well already release a minor version. Is this necessary?

yildirayunlu (approved):

packages/antd/src/hooks/form/useDrawerForm/index.spec.tsx
```diff
@@ -135,6 +135,10 @@ describe("useDrawerForm Hook", () => {
             result.current.formProps.onFinish?.({});
         });
 
+        await waitFor(() =>
+            expect(result.current.drawerProps.open).toBe(false),
+        );
+
         expect(result.current.drawerProps.open).toBe(false);
```
alicanerdurmaz, line 142: is this redundant?

```suggestion
        expect(result.current.drawerProps.open).toBe(false);
```

alicanerdurmaz (approved):

.changeset/forty-nails-remember.md
aliemir: This is for the fix of another issue, since we do `minor`, version will be bumped once but this changeset will be listed in patch changes.

packages/antd/src/hooks/form/useDrawerForm/index.spec.tsx
aliemir, line 142:
```suggestion
```

PR #4448, opened 2023-06-02
feat(core): make `useApiUrl` hook resource aware

BatuhanW:
Refactored `useApiUrl` to infer data provider from the current resource.

If you explicitly pass dataProviderName to useApiUrl hook, it will be used.
Then it will try to get current resource's dataProvider if it's defined.
Fallback to default data provider if none is found.

Screenshot 2023-06-03 at 02 03 18


documentation/docs/api-reference/core/hooks/data/useApiUrl.md
```diff
@@ -10,7 +10,8 @@ It is useful when you want to use the API URL in your custom hooks.
 
 ## Usage
 
-The `useApiUrl` hook does not expect any parameter. It will run the `getApiUrl` method from the `dataProvider` and return the result.
+The `useApiUrl` hook does not expect any parameter. It will run the `getApiUrl` method from the current resource's `dataProvider` and return the result.
```
salihozdemir, line 13: Actually `useApiUrl` hook can take a parameter that name is `dataProviderName`. When we say, "The `useApiUrl` hook does not expect any parameter." the user may think that the `useApiUrl` hook does not take any parameter. 🤔

packages/core/src/hooks/data/useApiUrl.spec.tsx
```diff
@@ -3,16 +3,110 @@ import { renderHook } from "@testing-library/react";
...
 
 describe("useApiUrl Hook", () => {
-    it("returns context value", async () => {
-        const { result } = renderHook(() => useApiUrl(), {
-            wrapper: TestWrapper({
-                dataProvider: MockJSONServer,
-                resources: [{ name: "posts" }],
-            }),
+    describe.each(["meta", "options"])(`should be called with %s`, (key) => {
+        describe("when dataProvider is *not* provided", () => {
```
salihozdemir, line 10:
```suggestion
        describe("when dataProvider is *not* provided", () => {
```
Can such a scenario happen?

packages/core/src/hooks/data/useApiUrl.spec.tsx
```diff
@@ -3,16 +3,110 @@ import { renderHook } from "@testing-library/react";
...
+
+            describe('when current resource has "dataProviderName" meta', () => {
+                it("returns resource's dataProvider", async () => {
+                    jest.spyOn(UseResource, "useResource").mockReturnValue({
+                        resource: {
+                            name: "categories",
+                            [key]: { dataProviderName: "categories" },
+                        },
+                        resources: [],
+                    });
```
salihozdemir, line 40: You can specify current resource, pathname etc. by using `mockRouterBindings` instead of spyOn.

salihozdemir (approved):

documentation/docs/api-reference/core/hooks/data/useApiUrl.md
BatuhanW: Good point, updated.

packages/core/src/hooks/data/useApiUrl.spec.tsx
BatuhanW: Here I mean `dataProviderName` param is undefined.

yildirayunlu (approved):

documentation/docs/api-reference/core/hooks/data/useApiUrl.md
```diff
@@ -35,6 +35,31 @@ const { data, isLoading } = useCustom<PostUniqueCheckResponse>({
...
+        >
+            {/* ... */}
+        </Refine>
+    );
+};
+    ...
+</Refine>
+
+
+const apiUrl = useApiUrl('other')
```
omeraplak, line 59:
```suggestion
const apiUrl = useApiUrl("other");
```

omeraplak (changes requested):

omeraplak (approved):

PR #4454, opened 2023-06-05
feat(mui): upgrade `@mui/x-data-grid` to `v6`

aliemir:
Updated the `@mui/x-data-grid` dependency to the latest version (`v6`). This update introduces some changes to the existing codebases which are addressed in Material UI's migration guide for `@mui/x-data-grid` from `v5` to `v6`. We've created a simple migration guide to navigate refine users through the changes that might be required in their codebases. While this guide does not cover all the changes, it will help you get started with the migration process.

**Breaking Changes**

`useDataGrid` no longer returns `page`, `pageSize`, `onPageChange` and `onPageSizeChange`. According to the changes in the `DataGrid` API, `useDataGrid` now returns `paginationModel` and `onPaginationModelChange` props which are used to control the pagination state of the `DataGrid` and contains the previous logic of `page`, `pageSize`, `onPageChange` and `onPageSizeChange`.

With this release, the peer dependency of `@mui/x-data-grid` is updated to `^6.6.0`.

### Self Check before Merge


omeraplak (approved):
🎖️

packages/mui/src/components/themedLayoutV2/index.tsx
```diff
@@ -26,12 +26,16 @@ export const ThemedLayoutV2: React.FC<RefineThemedLayoutV2Props> = ({
...
-                    }}
+                    sx={[
+                        {
+                            display: "flex",
+                            flexDirection: "column",
+                            flex: 1,
+                            minHeight: "100vh",
+                        },
+                        { overflow: "auto" },
+                        { overflow: "clip" },
```
salihozdemir, line 37: Don't they override each other with this syntax?

salihozdemir (approved):

packages/mui/src/components/themedLayoutV2/index.tsx
aliemir: they are both outputted like this:

```css
.some-class {
  overflow: auto;
  overflow: clip;
}
```

If `clip` is not supported, `auto` will be the fallback value.

yildirayunlu (approved):

alicanerdurmaz (approved):

PR #4498, opened 2023-06-10
Inferencer relationInfer issue with accessor key

### LINKED ISSUE #4497: [BUG] Inferencer relationInfer issue with accessor key
### Describe the bug

I override the field tags with field **_fieldTransformer_**

```tsx
    {
        key: "tags",
        type: "relation",
        relation: true,
        multiple: true,
        resource: { name: "tags", route: "/tags" },
        relationInfer: {
            accessor: "name",
        },
        accessor: "id",
    },
```

And If the record's tag array does not have any item, the next records on the list does not mapped with the accessor key.

```json
{
  "id": 1,
  "title": "1111123dasdasd",
  "content": "adasdasdasd",
  "published": true,
  "authorId": null,
  "categoryId": 2,
  "tags": [
    
  ],
}

```

Screenshot 2023-06-10 at 14 29 54

Also on the show page, it says "not-handled - relation with multiple but no resource"

Screenshot 2023-06-10 at 14 32 00

### Steps To Reproduce

To solve this I have made some changes to the file while debugging

.../packages/inferencer/src/inferencers/mui/show.tsx

.../packages/inferencer/src/use-relation-fetch/index.ts

### Expected behavior

When the array of a key does not have any item, it should show empty view.

### Screenshot

_No response_

### Desktop

_No response_

### Mobile

- UI Framework: MUI

### Additional Context

_No response_

rasitcolakel:
Inferencer relationInfer issue with accessor key when the field is multiple but not have any object/value in the array

### Closing issues

closes #4497 


BatuhanW:
Thanks for the PR @rasitcolakel, probably we have the same situtation for other inferencers too (antd, chakra-ui), could you check them if you feel like?

rasitcolakel:
> Thanks for the PR @rasitcolakel! Probably we have the same situtation for other inferencers too (antd, chakra-ui), could you check them if you feel like?

Hi @BatuhanW , I checked the other inferencers and pushed a new commit.

Also, If the user passes the "relationInfer" parameter to the "fieldTransformer", the relationInfer is set without being predicted.

aliemir (changes requested):
Hey @rasitcolakel, thank you for the contribution! Sorry for the late review 🙏 

I understood the change in the `useRelationFetch` hook but having hard time understanding the context of the changes in the `show` inferencers. If the array inference is correct, isn't always outputting the map function a better choice for supporting the future changes in the record? If there's no specific case for these changes, I think we can revert them to keep using map only, imho. 🤔

aliemir:
BTW, can you also add a changeset for the changes for the release? 🙏 You can check out our contributing docs to see how: https://refine.dev/docs/contributing/#changeset

rasitcolakel:
> Hey @rasitcolakel, thank you for the contribution! Sorry for the late review 🙏
> 
> I understood the change in the `useRelationFetch` hook but having hard time understanding the context of the changes in the `show` inferencers. If the array inference is correct, isn't always outputting the map function a better choice for supporting the future changes in the record? If there's no specific case for these changes, I think we can revert them to keep using map only, imho. 🤔


The problem with the `show` inferencer is when a record has a property that is an array of a relation like the tags below. It does not check whether the property has any data. If the array is empty, it sends a request to `${API_URL}/api/tags?`, and it gets all tags, and renders all of them as you can see the first picture.

```json
// http://localhost:3000/api/posts/11
{
  "id": 11,
  "title": "123456",
  "content": "12313",
  "published": false,
  "authorId": 5,
  "categoryId": 2,
  "tags": [
    
  ],
  "likes": [
    {
      "id": 3,
      "postId": 11,
      "authorId": 12
    }
  ]
}
```

Picture one
Screenshot 2023-06-19 at 17 25 04


After, changing the `show` inferencer, it renders like the picture two. That is why I updated the `show` inferencers of UI frameworks and headless UI 

Screenshot 2023-06-19 at 17 28 31

rasitcolakel:
> BTW, can you also add a changeset for the changes for the release? 🙏 You can check out our contributing docs to see how: https://refine.dev/docs/contributing/#changeset

Thanks, I will look at it :)

aliemir:
Discussed with @rasitcolakel about the changes in the show components, found a better approach to the scenario. @rasitcolakel will implement the changes and update the PR. 🚀 🚀 🙏

PR #4523, opened 2023-06-13
feat: add `useLoadingOvertime` hook

yildirayunlu:
if you need to do something when the loading time exceeds the specified time, refine provides the `useLoadingOvertime` hook. It returns the elapsed time in milliseconds.

```tsx
const { elapsedTime } = useLoadingOvertime({
   isLoading,
   interval: 5000,
   onInterval(elapsedInterval) {
       console.log("loading overtime", elapsedInterval);
   },
});
```

### Test plan (required)

image

### Self Check before Merge


alicanerdurmaz (approved):

examples/base-antd/src/App.tsx
```diff
@@ -16,9 +16,32 @@ import { ConfigProvider } from "antd";
...
 
 import { PostList, PostCreate, PostEdit, PostShow } from "../src/pages/posts";
+import { useState } from "react";
 
 const API_URL = "https://api.fake-rest.refine.dev";
 
+// TODO: Add for demo, remove later
+const DashboardPage = () => {
+    const [isLoading, setIsLoading] = useState(false);
+    const { elapsedTime } = useLoadingOvertime({
```
salihozdemir, line 26: Will you create a separate example for this feature?

.changeset/weak-fishes-wave.md
```diff
@@ -0,0 +1,17 @@
...
+if you need to do something when the loading time exceeds the specified time, refine provides the `useLoadingOvertime` hook. It returns the elapsed time in milliseconds.
+
+```tsx
+const { elapsedTime } = useLoadingOvertime({
+   isLoading,
+   interval: 5000,
+   onInterval(elapsedInterval, context) {
+       console.log("loading overtime", elapsedInterval, context);
+   },
+});
```
salihozdemir, line 16: In my opinion, we should show the `overtime` usage on `<Refine/>` component.

packages/core/src/hooks/useLoadingOvertime/index.ts
```diff
@@ -0,0 +1,100 @@
...
+ * });
+ */
+export const useLoadingOvertime = ({
+    isLoading,
+    interval: intervalProp,
+    onInterval: onIntervalProp,
+}: UseLoadingOvertimeProps): UseLoadingOvertimeReturnType => {
+    const resourceContextData = useResource();
+    const [elapsedTime, setElapsedTime] = useState<number | undefined>(
+        undefined,
```
salihozdemir, line 57: I think it would be better if the default value is 0
```suggestion
        0,
```

packages/core/src/hooks/useLoadingOvertime/index.spec.tsx
```diff
@@ -0,0 +1,167 @@
...
+        act(() => {
+            // default 2000
+            jest.advanceTimersByTime(3000);
+        });
+
+        const { elapsedTime } = result.current;
+        expect(elapsedTime).toBe(2000);
+    });
+
+    it("should override global interval", () => {
```
salihozdemir, line 68: We can also add a test where we define a global interval as "should apply global interval if interval doesn't provided".

yildirayunlu: I added a test with `global interval and onInterval run`.

examples/base-antd/src/App.tsx
yildirayunlu: A separate example can be made when we integrate it into data hooks. Just for demo for now.

packages/core/src/contexts/refine/IRefineContext.ts
```diff
@@ -42,6 +44,7 @@ export interface IRefineContextOptions {
         afterClone: RedirectAction;
         afterEdit: RedirectAction;
     };
+    overtime?: Omit<UseLoadingOvertimeProps, "isLoading">;
```
BatuhanW, line 47: AFAIK, we are providing a default value for `interval`, so it's not optional, this type is wrong because `interval` is optional. We should mark is as required. Some pseudo code for this:

```
Required<Pick<UseLoadingOvertimeProps, "interval">> & Omit<UseLoadingOvertimeProps, "isLoading", "interval">

```

packages/core/src/definitions/helpers/handleRefineOptions/index.spec.ts
```diff
@@ -50,6 +50,9 @@ describe("handleRefineOptions", () => {
                 afterEdit: "show",
             },
             breadcrumb: false,
+            overtime: {
+                interval: 1000,
```
BatuhanW, line 54: Are they needed since we already have a default value?

packages/core/src/definitions/helpers/handleRefineOptions/index.spec.ts
```diff
@@ -148,6 +157,9 @@ describe("handleRefineOptions", () => {
                 afterCreate: "list",
                 afterEdit: "list",
             },
+            overtime: {
+                interval: 1000,
```
BatuhanW, line 161: Maybe we can add assertion about `interval` having a default value of 1000.

packages/core/src/hooks/useLoadingOvertime/index.ts
```diff
@@ -0,0 +1,100 @@
+import { useEffect, useState } from "react";
+import { UseResourceReturnType, useRefineContext, useResource } from "..";
+
+export type IUseLoadingOvertimeOnIntervalContext = UseResourceReturnType;
```
BatuhanW, line 4: I think it's too much. We shouldn't return all of the resources array in the context, we better return only relative `resource` object.

packages/core/src/hooks/useLoadingOvertime/index.ts
```diff
@@ -0,0 +1,100 @@
...
+    const [elapsedTime, setElapsedTime] = useState<number | undefined>(
+        undefined,
+    );
+    // get overtime options from refine context
+    const { options } = useRefineContext();
+    const { overtime } = options;
+
+    // pick props or refine context options
+    const interval = intervalProp ?? overtime?.interval;
+    const onInterval = onIntervalProp ?? overtime?.onInterval;
```
BatuhanW, line 65: Related to the review before, this shouldn't be optional, we should have value for sure.

BatuhanW (changes requested):

packages/core/src/contexts/refine/IRefineContext.ts
yildirayunlu: @BatuhanW  True, if `overtime` is defined then `interval` should be required.

packages/core/src/definitions/helpers/handleRefineOptions/index.spec.ts
yildirayunlu: I think there is. Because here we are already checking the default values.

packages/core/src/definitions/helpers/handleRefineOptions/index.spec.ts
yildirayunlu: Here we do this check.

packages/core/src/hooks/useLoadingOvertime/index.ts
yildirayunlu: fixed `interval` is not optional

packages/core/src/hooks/useLoadingOvertime/index.ts
yildirayunlu: Only passes `resource` object

alicanerdurmaz (approved):

BatuhanW (approved):

aliemir (changes requested):
In the `onInterval` callback, we're passing a `context` that is not necessarily related to the parent hook since it's only inferred from the router. 

Since this is used in our hooks, we should accept a `context` value (probably in `Record` type) and pass the values we can provide (such as `resource`, `filters`, `meta` in `useList` and `values`, `resource`, `meta` etc. in `useCreate` and similar.)

This way we can give the user a better understanding in `onInterval` callback. (At least a correct one) 

Example: 

```tsx
const overtime = useLoadingOvertime({
  context: { meta, resource, values },
});
```

packages/core/src/contexts/refine/IRefineContext.ts
```diff
@@ -29,8 +29,8 @@ export interface IRefineOptions {
         clientConfig?: QueryClientConfig | InstanceType<typeof QueryClient>;
         devtoolConfig?: React.ComponentProps<typeof ReactQueryDevtools> | false;
     };
-    overtime?: Omit<UseLoadingOvertimeProps, "isLoading" | "interval"> &
-        Required<Pick<UseLoadingOvertimeProps, "interval">>;
+    overtime?: Omit<UseLoadingOvertimeCoreProps, "isLoading" | "interval"> &
+        Required<Pick<UseLoadingOvertimeCoreProps, "interval">>;
```
BatuhanW, line 33: Maybe we can move this type definition to `packages/core/src/hooks/useLoadingOvertime/index.ts` and use it from there.

aliemir (approved):

aliemir (changes requested):
In the documentation, can we also add `overtime` to the return values? 💯

documentation/docs/api-reference/core/hooks/data/useCreateMany/index.md
```diff
@@ -276,6 +299,16 @@ Returns an object with TanStack Query's `useMutation` return values.
 
 > For more information, refer to the [`useMutation` documentation &#8594](https://tanstack.com/query/v4/docs/react/reference/useMutation)
 
+### Additional Values
+#### `overtime`
+
+`overtime` object is returned from this hook. `elapsedTime` is the elapsed time in milliseconds. It becomes `undefined` when the request is completed.
+
+```tsx
+const { overtime } = useCreate();
```
aliemir, line 308:
```suggestion
const { overtime } = useCreateMany();
```

Other hooks also uses the same snippet with `useCreate`, can we change them accordingly?

aliemir (changes requested):
Noticed some inconsistencies in the documentation. Waiting for their updates and we're good to go!

salihozdemir:
I didn't find the `useLoadingOvertime` hook. Am I missing something?

aliemir:
> I didn't find the `useLoadingOvertime` hook. Am I missing something?

We don't have a separate documentation for that and kept it internal for now.

aliemir (approved):

PR #4527, opened 2023-06-14
feat: support multiple `resource` usage with the same name via the `identifier`

salihozdemir:
Previously our primitive data hooks(`useList`, `useCreate`, etc.) only worked with resource name. So if you had multiple `resource` usage with the same name, it would cause issues. Now the hooks support `identifier` to distinguish between the resources.

### Test plan (required)

The new tests are added.

image

### Self Check before Merge


yildirayunlu (approved):

alicanerdurmaz (approved):

.changeset/early-fireants-accept.md
```diff
@@ -0,0 +1,7 @@
+---
+"@refinedev/core": minor
+---
+
+fix: support multiple `resource` usage with the same name via the `identifier`
+
+Previously our primitive data hooks(`useList`, `useCreate`, etc.) only worked with resource name. So if you had multiple `resource` usage with the same name, it would cause issues. Now the hooks support `identifier` to distinguish between the resources.
```
BatuhanW, line 7: Even though the term `primitive data hooks` is clear for us, it may not be clear for everyone, let's make a list of these refactored hooks and add a sentence like, `following hooks now respects identifier` etc..

packages/antd/src/components/buttons/delete/index.tsx
```diff
@@ -59,6 +59,7 @@ export const DeleteButton: React.FC<DeleteButtonProps> = ({
     const { id, resource } = useResource(
         resourceNameFromProps ?? propResourceNameOrRouteName,
     );
+    const resourceIdentifierOrName = resource?.identifier ?? resource?.name;
```
BatuhanW, line 62: How do you feel about moving this one into the `useResource` hook itself, since it's repetitive everywhere?

Let's expose `identifier` key from useResource hook and use it everywhere.

```tsx
const { identifier } = useResource(...)
```

packages/core/src/components/canAccess/index.tsx
```diff
@@ -54,95 +41,37 @@ export type CanAccessProps = CanAccessBaseProps | CanAccessWithoutParamsProps;
...
-                }
-            } else {
-                /** If `resource` is passed as an IResourceItem, use it or `resource` is undefined and cannot be inferred. */
-                resource = resourceFromRouter;
-            }
+        resource,
+        id: idFromRoute,
+        action: actionFromRoute,
+    } = useResource(resourceFromProp);
+    const { resource: inferedResource } = useResource();
```
BatuhanW, line 55:
```suggestion
    const { resource: inferredResource } = useResource();
```

packages/core/src/components/canAccess/index.tsx
```diff
@@ -54,95 +41,37 @@ export type CanAccessProps = CanAccessBaseProps | CanAccessWithoutParamsProps;
...
+
+    const getDefaultId = () => {
+        const idFromPropsOrRoute = params?.id ?? idFromRoute;
+
+        if (
+            resourceFromProp &&
+            resourceFromProp !== inferedResource?.identifier &&
+            resourceFromProp !== inferedResource?.name
```
BatuhanW, line 63: Once we export `identifier` from `useResource`, we can simplify this condition:

```
resourceFromProp && resourceFromProp !== identifier
```

BatuhanW (changes requested):

packages/core/src/hooks/data/useCreateMany.ts
```diff
@@ -102,23 +102,26 @@ export const useCreateMany = <
         useCreateManyParams<TData, TError, TVariables>
     >(
         ({
-            resource,
+            resource: resourceName,
```
BatuhanW, line 105: 🥇

packages/core/src/hooks/data/useCreateMany.ts
```diff
@@ -102,23 +102,26 @@ export const useCreateMany = <
...
 
             const selectedDataProvider = dataProvider(
-                pickDataProvider(resource, dataProviderName, resources),
+                pickDataProvider(identifier, dataProviderName, resources),
             );
 
             if (selectedDataProvider.createMany) {
                 return selectedDataProvider.createMany<TData, TVariables>({
-                    resource,
+                    resource: resource.name,
```
BatuhanW, line 124: Shouldn't we use identifier here too?

packages/chakra-ui/src/components/crud/list/index.tsx
```diff
@@ -95,12 +93,12 @@ export const List: React.FC<ListProps> = (props) => {
                 className={RefinePageHeaderClassNames.Title}
             >
                 {translate(
-                    `${resource?.name}.titles.list`,
+                    `${identifier}.titles.list`,
```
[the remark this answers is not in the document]
salihozdemir: Added the changesets 👍

packages/core/src/components/canAccess/index.tsx
```diff
@@ -54,95 +41,33 @@ export type CanAccessProps = CanAccessBaseProps | CanAccessWithoutParamsProps;
...
+            return params?.id;
         }
-    }
+
+        return idFromPropsOrRoute;
+    };
+    const defaultId = getDefaultId();
 
     const { data } = useCan({
         resource: resourceFromProp ?? resource?.name,
```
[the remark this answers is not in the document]
BatuhanW: Just to give context on decision, we are using identifier internally to find resource, communication with outside providers should be through name.

BatuhanW (changes requested):
Let's add a warning about possible i18n issue, since we are using `identifier` now, instead of resource name.

BatuhanW (approved):

yildirayunlu (approved):

packages/core/src/hooks/form/useForm.ts
```diff
@@ -479,7 +411,7 @@ export const useForm = <
                 return mutateCreate(
                     {
                         values,
-                        resource: resource.name,
+                        resource: identifier ?? resource.name,
```
aliemir, line 414: As I can see, `identifier` satisfies both the `identifier` value and the `name`. Do we need the extra `?? resource.name` here? 🤔 Or do we have a type issue that requires us to do this? @salihozdemir

packages/core/src/hooks/form/useForm.ts
```diff
@@ -519,7 +451,7 @@ export const useForm = <
         const variables: UpdateParams<TResponse, TResponseError, TVariables> = {
             id: id ?? "",
             values,
-            resource: resource.name,
+            resource: identifier ?? resource.name,
```
aliemir, line 454: Same as above

packages/core/src/hooks/form/useForm.ts
salihozdemir: Actually, this is not a type problem. If `useResource` takes a parameter, `identifier` is always a string. However, if `useResource` takes a value `undefined | string`, there is a possibility that `identifier` is undefined. So, we need to add `??` in here.

packages/core/src/hooks/form/useForm.ts
salihozdemir: Same answer above :)

aliemir (approved):

packages/core/src/hooks/form/useForm.ts
aliemir: Thank you

PR #4539, opened 2023-06-16
docs: add `identifier` usage and more information under the `resource` sections

salihozdemir:
- Added the `identifier` usage.
- Added more info to `resource` usage.
- Added the `select` doc on `useResource` hook.


alicanerdurmaz (approved):

documentation/docs/api-reference/core/components/refine-config.md
```diff
@@ -148,7 +148,62 @@ https://api.fake-rest.refine.dev/posts/1
 
 ### `identifier`
 
-You can pass this value to a resource and it will be used as the main matching key for the resource. This is useful when you have multiple resources with the same `name`.
+The `identifier` value servers as the main matching key for a resource. It allows you to effectively differentiate between multiple resources that share the same `name`.
```
BatuhanW, line 151:
```suggestion
The `identifier` value serves as the main matching key for a resource. It allows you to effectively differentiate between multiple resources that share the same `name`.
```

documentation/docs/api-reference/core/components/refine-config.md
```diff
@@ -148,7 +148,62 @@ https://api.fake-rest.refine.dev/posts/1
...
+
+There are scenarios where you may require multiple resources with the same `name` but distinct `meta` values. For instance, you might want a posts resource utilizing the default data provider and another posts resource utilizing the "typicode" data provider. In this case, you can use the `identifier` to differentiate between them.
```
BatuhanW, line 153:
```suggestion
There are scenarios where you may have multiple resources with the same `name` but different `meta` values. For instance, you might want a `posts` resource utilizing the default data provider and another `posts` resource utilizing the "typicode" data provider. In this case, you can use the `identifier` to differentiate between them.
```

PR #4552, opened 2023-06-19
fix(ci): every chunk doesnt have equal amount of e2e specs

BatuhanW:
Before this PR, most of the e2e specs are running at chunks `1,4 and 6`, and other chunks are finishing quickly. Since most of the specs are accumulated in chunk 4, we are wasting a lot of time while waiting for chunk 4 to finish.

With this update, every chunk more or less will have same amount of e2e tests.


BatuhanW:
Closing in favor of https://github.com/refinedev/refine/pull/4467, we'll release this change within linked PR.

PR #4558, opened 2023-06-20
feat: add `<LoadingOvertime/>` component

salihozdemir:
**`<LoadingOvertimeIndicator/>`**

 A component that renders children and overtime components based on the elapsed time.
 The elapsed time is calculated using the `useLoadingOvertime` hook.
 If the `elapsedTime` prop is provided, the component will not use the `useLoadingOvertime` hook.

**Example Usage From Core**

```tsx
import { LoadingOvertimeIndicator }  from "@refinedev/core";

const overtimeComponents = {
    4000: <div>It's taking a bit longer than expected.</div>,
    8000: <div>This is taking longer than expected, please hang on.</div>,
};

<LoadingOvertimeIndicator
    overtimeComponents={overtimeComponents}
    elapsedTime={4000}
>
    <div>content</div>
</LoadingOvertimeIndicator>,
```

### Test plan (required)

- [x] Add component tests
- [x] Add shared tests to `@refinedev/ui-tests`
- [x] Use the shared tests on UI packages

### Self Check before Merge

https://refinedev.atlassian.net/browse/REF-10


packages/antd/src/components/loadingOvertimeIndicator/index.tsx
```diff
@@ -0,0 +1,52 @@
+import React from "react";
+import { LoadingOvertimeIndicator, useTranslate } from "@refinedev/core";
+import { Alert } from "antd";
+
+type AlertComponentProps = {
+    translateKey: string;
+    message: string;
+};
+
+const AlertComponent = ({ translateKey, message }: AlertComponentProps) => {
```
BatuhanW, line 10: I think accepting messages for different intervals would be nice. So, users can just use antd component, with their messages.

salihozdemir: We are not exporting this component. Instead, we export the `overtimeComponent`

yildirayunlu (approved):

alicanerdurmaz (approved):

PR #4568, opened 2023-06-21
feat: add the `textTransformers` option to `<Refine/>` component

salihozdemir:
By default, **refine** uses the following packages to transform some strings:
- [plurize](https://www.npmjs.com/package/pluralize)
- [humanize-string](https://www.npmjs.com/package/humanize-string)

In some languages other than English, these transformations could cause problems. So, this PR provides a solution for this.

``` tsx
<Refine options={{
    textTransformers: {
          humanize: (word) => handleHumanize(word),
          plural: (word) => handlePlurize(word),
          singular: (word) => handleSingularize(word),
      }
  }} 
/>
```

### Test plan (required)

WIP:

### Self Check before Merge


documentation/docs/api-reference/core/components/refine-config.md
```diff
@@ -693,6 +693,39 @@ const App: React.FC = () => (
...
+The `textTransformers` option in **refine** is used to transform the resource name displayed on the user interface (UI). By default, if you define a resource named `posts`, refine will display it as `Posts`. Similarly, when you delete a record, notification messages will be shown as `Post deleted successfully.`.
+
+You have the flexibility to customize these messages by using the `textTransformers` option. For instance, if you wish to disable any transformation, you can set the `textTransformers` option as shown in the example below:
+
+```tsx
+const App: React.FC = () => (
+    <Refine
+        ...
+        options={{
+            textTransformers: {
```
BatuhanW, line 707: I think it would be much better to have `disabled` boolean option, defaults to false.

So users can easier disable all transformations, without defining all the functions repetitively.

documentation/docs/api-reference/core/components/refine-config.md
```diff
@@ -693,6 +693,39 @@ const App: React.FC = () => (
...
+### `textTransformers`
+
+The `textTransformers` option in **refine** is used to transform the resource name displayed on the user interface (UI). By default, if you define a resource named `posts`, refine will display it as `Posts`. Similarly, when you delete a record, notification messages will be shown as `Post deleted successfully.`.
+
+You have the flexibility to customize these messages by using the `textTransformers` option. For instance, if you wish to disable any transformation, you can set the `textTransformers` option as shown in the example below:
+
+```tsx
+const App: React.FC = () => (
+    <Refine
+        ...
```
BatuhanW, line 705:
```suggestion
        // ...
```

.changeset/lucky-dingos-roll.md
```diff
@@ -0,0 +1,24 @@
...
+feat: add the `textTransformers` option to `<Refine/>` component
+
+The `textTransformers` option in **refine** is used to transform the resource name displayed on the user interface (UI). By default, if you define a resource named `posts`, refine will display it as `Posts`. Similarly, when you delete a record, notification messages will be shown as `Post deleted successfully.`.
+
+You have the flexibility to customize these messages by using the `textTransformers` option. For instance, if you wish to disable any transformation, you can set the `textTransformers` option as shown in the example below:
+
+```tsx
+const App: React.FC = () => (
+    <Refine
+        ...
```
BatuhanW, line 14:
```suggestion
        // ...
```

Without making it comment, formatting doesn't work for codeblocks. I think it's nice to have formatting in these code-blocks, so we can understand if we've made a typo in general.

packages/core/src/hooks/breadcrumb/index.ts
```diff
@@ -35,10 +35,11 @@ export const useBreadcrumb = ({
     const routerType = useRouterType();
     const { i18nProvider } = useContext(TranslationContext);
     const parsed = useParsed();
-
     const translate = useTranslate();
-
     const { resources, resource, action } = useResource();
+    const { options } = useRefineContext();
+
+    const { textTransformers } = options;
```
BatuhanW, line 42: 😄  Nitpick: 

```suggestion
    const { options: { textTransformers } } = useRefineContext();
```

documentation/docs/api-reference/core/components/refine-config.md
salihozdemir: @aliemir What do you think about it?

aliemir: These transformations are already used in the core package and the ability we provide with this PR is not for a customization on the raw values but an extension to the transformations. I'm not sure if it will be anywhere useful to disable them.

packages/react-router-v6/src/document-title-handler.tsx
```diff
@@ -30,11 +29,9 @@ export function DocumentTitleHandler({ handler }: Props) {
...
         resource?.label ??
         resource?.meta?.label ??
-        capitalize(
-            getUserFriendlyName(
-                identifier,
-                action === "list" ? "plural" : "singular",
-            ),
+        getUserFriendlyName(
+            identifier,
+            action === "list" ? "plural" : "singular",
```
BatuhanW: This change didn't fail any spec. Let's add some tests.

BatuhanW (changes requested):

packages/react-router-v6/src/document-title-handler.tsx
salihozdemir: We don't have any test environment in router packages. So, we didn't test it

salihozdemir: Also, this change does not cause any perpetrators. Because `getUserFriendlyName` already does `capitalize`.

BatuhanW (approved):

PR #4569, opened 2023-06-21
fix(examples): inferencer e2e wait category intercept

alicanerdurmaz:
The inferencer tests were not passing because it couldn't find the getCategories intercept. getCategories intercept deleted because it is not needed

### Self Check before Merge


examples/inferencer-material-ui/cypress/e2e/all.cy.ts
```diff
@@ -44,7 +44,6 @@ describe("inferencer-material-ui", () => {
 
     it("should show resource", () => {
         cy.interceptGETBlogPost();
```
BatuhanW, line 46: Isn't this duplicated? I think we also have it at the top, before block.

BatuhanW (approved):

PR #4599, opened 2023-06-25
fix(core): handle missing translation issue in document titles

aliemir:
Update default document title generation to use the fallback title when `i18n`'s `translate` function returns the `key` value.

### Test plan (required)

```sh
Test Suites: 126 passed, 126 total
Tests:       1001 passed, 1001 total
Snapshots:   18 passed, 18 total
Time:        82.581 s, estimated 90 s
Ran all test suites.
```

### Self Check before Merge


packages/core/src/definitions/helpers/safe-translate/index.test.ts
```diff
@@ -0,0 +1,31 @@
+import { safeTranslate } from ".";
```
BatuhanW, line 1: I think it's better to follow naming convention for all other tests(except inferencer tests 😄)

`index.spec.ts`

packages/core/src/definitions/helpers/safe-translate/index.ts
```diff
@@ -0,0 +1,20 @@
+import { useTranslate } from "@hooks/translate";
```
BatuhanW, line 1: Can we use like below?
```suggestion
import type { useTranslate } from "@hooks/translate";
```

packages/core/src/definitions/helpers/safe-translate/index.test.ts
aliemir: Done ✅

packages/core/src/definitions/helpers/safe-translate/index.ts
aliemir: Not needed actually, but updated them anyways.

BatuhanW (approved):

PR #4740, opened 2023-07-26
fix(core): telemetry call fallback

aliemir:
Use `fetch` for telemetry calls as a fallback for `Image` when it's not available. This fixes an issue where telemetry calls would fail in some environments.

To make sure `fetch` is called only once, added a ref check before the effect.

### Self Check before Merge

### DIFF
--- a/packages/core/src/components/telemetry/index.tsx
+++ b/packages/core/src/components/telemetry/index.tsx
@@ -4,28 +4,49 @@ import { useTelemetryData } from "@hooks/useTelemetryData";
 
 import { ITelemetryData } from "../../interfaces/telemetry";
 
-const encode = (payload: ITelemetryData): string => {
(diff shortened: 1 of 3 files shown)


packages/core/src/components/telemetry/index.tsx
```diff
@@ -1,31 +1,57 @@
-import { useEffect } from "react";
+import React from "react";
 
 import { useTelemetryData } from "@hooks/useTelemetryData";
 
 import { ITelemetryData } from "../../interfaces/telemetry";
 
-const encode = (payload: ITelemetryData): string => {
-    const stringifyedPayload = JSON.stringify(payload || {});
+const encode = (payload: ITelemetryData): string | undefined => {
```
omeraplak, line 7: Could we add unit tests?

omeraplak (approved):

PR #4772, opened 2023-08-02
fix(antd):  `useModalForm` and `useDrawerForm` sends request twice when `syncWithLocation` is true

alicanerdurmaz:
If `syncWithLocation` is enabled, after the modal is opened, `useModalForm` and `useDrawerForm` was making another request because the URL was changing. This issue has been resolved by removing the URL parameters that indicate the modal is open from the `queryKey`.

### Test plan (required)

- e2e tests are updated.
- new unit tests are added.

image

### Self Check before Merge


yildirayunlu (approved):

packages/antd/src/hooks/form/useDrawerForm/index.spec.tsx
```diff
@@ -265,4 +265,39 @@ describe("useDrawerForm Hook", () => {
...
+            },
+        );
+
+        await act(async () => {
+            result.current.show();
+        });
+
+        await waitFor(() => expect(result.current.drawerProps.open).toBe(true));
+
+        await waitFor(() => expect(mockGetOne.mock.calls.length).toBe(1));
```
salihozdemir, line 298:
```suggestion
        await waitFor(() => expect(mockGetOne).toBeCalledTimes(1));
```

packages/antd/src/hooks/form/useDrawerForm/index.spec.tsx
```diff
@@ -265,4 +265,39 @@ describe("useDrawerForm Hook", () => {
...
+        expect(mockGetOne.mock.calls[0][0].meta?.["drawer-posts-edit"]).toBe(
+            undefined,
+        );
+    });
```
salihozdemir, line 302:
```suggestion
        expect(mockGetOne.toBeCalledWith(
            expect.objectContaining({
            meta:   expect.not.objectContaining({
            "drawer-posts-edit": expect.any()
            )}
             })
        );
    });
```

The suggestion may won't work however you can try like this

salihozdemir (changes requested):

salihozdemir (approved):

PR #4804, opened 2023-08-08
test: add `projectId` tests on `refineOptions`

yildirayunlu:
Add `projectId` tests on `refineOptions`

### Self Check before Merge


packages/core/src/hooks/useTelemetryData/index.spec.ts
```diff
@@ -292,4 +294,41 @@ describe("useTelemetryData Hook", () => {
             expect(providers.router).toBeTruthy();
         });
     });
+
+    describe("projectId", () => {
+        const mockRefineProvider: IRefineContextProvider = {
+            hasDashboard: false,
```
omeraplak, line 300: Is the 'hasDashboard' line necessary for this test?

yildirayunlu: `hasDashboard` is required. I added, because it is not from `defaultRefineOptions`

PR #4847, opened 2023-08-22
feat: crm contact/list page

yildirayunlu:
Contact List screen of the created CRM application.

### Self Check before Merge


examples/app-crm/src/routes/contacts/wrapper/index.tsx
```diff
@@ -0,0 +1,296 @@
...
+                </Row>
+
+                <Pagination
+                    style={{ textAlign: "end", marginTop: "1rem" }}
+                    total={data?.total}
+                    onChange={(page, pageSize) => {
+                        setCurrent(page);
+                        setPageSize(pageSize);
+                    }}
+                />
```
aliemir, line 255: We can directly use `tableProps.pagination` props here without setting them explicitly in the code.

examples/app-crm/src/routes/contacts/wrapper/index.tsx
```diff
@@ -0,0 +1,313 @@
...
+                                },
+                                {
+                                    label: "Won",
+                                    value: ContactStatus.WON,
+                                },
+                                {
+                                    label: "Churned",
+                                    value: ContactStatus.CHURNED,
+                                },
+                            ]}
```
aliemir, line 131: We can move this array outside of the component like:

```tsx
const statusOptions = Object.keys(ContactStatus).map((key) => ({
    label: `${key[0]}${key.slice(1).toLowerCase()}`,
    value: ContactStatus[key as keyof typeof ContactStatus],
}));
```

and pass `options={statusOptions}`

🤔

aliemir (approved):

PR #4906, opened 2023-09-04
feat(core): add `onUnauthorized` callback to `<CanAccess />` component

alicanerdurmaz:
feat: added `onUnauthorized` callback to `<CanAccess />` component. This callback to be called when [`useCan`](https://refine.dev/docs/api-reference/core/hooks/accessControl/useCan/) returns false.

```tsx
<CanAccess
    onUnauthorized={({ resource, reason, action, params }) =>
        console.log(
            `You cannot access ${resource}-${params.id} resource with ${action} action because ${reason}`,
        )
    }
>
    <YourComponent />
</CanAccess>
```

### Test plan (required)

image

### Self Check before Merge


aliemir (approved):

packages/core/src/components/canAccess/index.spec.tsx
```diff
@@ -57,6 +74,20 @@ describe("CanAccess Component", () => {
             expect(container).toBeTruthy();
             expect(queryByText("Accessible")).not.toBeInTheDocument();
         });
+
+        await waitFor(() => {
+            expect(onUnauthorized).toHaveBeenCalledWith({
```
BatuhanW, line 79: Can we also test it's triggered just once, when the result of the `can` changes` ?

packages/core/src/components/canAccess/index.tsx
```diff
@@ -55,25 +69,47 @@ export const CanAccess: React.FC<CanAccessProps> = ({
...
+                params,
+            });
+        }
+    }, [
+        onUnauthorized,
+        resourceName,
+        action,
+        data?.can,
+        params.id,
+        params.resource,
```
BatuhanW, line 110: Do we need all of these dependencies? I think just `data?.can` could be sufficient.

packages/core/src/components/canAccess/index.tsx
```diff
@@ -25,6 +25,19 @@ type CanAccessBaseProps = {
...
+    onUnauthorized?: (props: {
+        resource?: string;
+        reason?: string;
+        action: string;
+        params: {
+            resource?: IResourceItem & { children?: ITreeMenu[] };
+            id?: BaseKey;
+            [key: string]: any;
+        };
+    }) => void;
```
BatuhanW, line 40:
```suggestion
    }) => Promise<void> | void;
```

I think we should allow it to be an async function, even though we won't await it.

packages/core/src/components/canAccess/index.tsx
```diff
@@ -25,6 +25,19 @@ type CanAccessBaseProps = {
      * Content to show if access control returns `false`
      */
     fallback?: React.ReactNode;
+    /**
+     * Callback function to be called if access control returns `false`
+     */
+    onUnauthorized?: (props: {
```
BatuhanW, line 31: Can we make `props` an interface, instead of writing them inline?

```suggestion
    onUnauthorized?: (props: OnUnauthorizedProps)
```

alicanerdurmaz: 🚀

packages/core/src/components/canAccess/index.tsx
```diff
@@ -1,8 +1,21 @@
-import React from "react";
+import React, { useEffect } from "react";
 
 import { useCan, useResource } from "@hooks";
 import { BaseKey, IResourceItem, ITreeMenu } from "../../interfaces";
 
+type Params = {
```
BatuhanW, line 6: I think this name is too generic. Optionally, we can rename it to something more specific, so it will be easier to read.

BatuhanW (approved):
LGTM 👍🏼 

Added a minor comment.

BatuhanW (approved):

salihozdemir (approved):

PR #4911, opened 2023-09-05
feat: Add module federation example

BatuhanW:
Added a module federation example using vite.


documentation/docs/examples/module-federation/vite.md
```diff
@@ -0,0 +1,58 @@
+---
+id: vite
+title: Module Federation with Vite
+example-tags: [module-federation, vite, microfrontend]
```
yildirayunlu, line 4:
```suggestion
example-tags: [module-federation]
```
I think just `module-federation` will be enough, because we usually use only 1 tag in the examples.

yildirayunlu (approved):

documentation/docs/examples/module-federation/vite.md
```diff
@@ -0,0 +1,58 @@
+---
+id: vite
+title: Module Federation with Vite
```
BatuhanW, line 3:
```suggestion
title: Module Federation
```

PR #4945, opened 2023-09-15
fix: antd notificationProvider bug

### LINKED ISSUE #4654: [BUG] Antd notificationProvider notifications don't follow theme style
### Describe the bug

Notifications have a white background even tho the dark theme is set.

### Steps To Reproduce

1. Go to [customization-theme-antd](https://codesandbox.io/embed/github/refinedev/refine/tree/next/examples/) example.
2. Set the dark theme by clicking the moon icon.
3. Click the edit button of any post, then click Save.
4. Notice how the "Successfully updated post" notification is white even tho the dark theme is set.

### Expected behavior

Notifications should follow the theme style, like they do in the [ant design](https://ant.design/components/notification?theme=dark) examples.

### Desktop

_No response_

### Mobile

_No response_

### Additional Context

_No response_

MahirMahdi:

`notificationProvider` cannot access the theme context when the static `notification` method is used. So used the `App` component's `notification` instance to access the theme context. 

Passed the `App` component as a child of `ConfigProvider` to access the theme context.

### Test plan (required)

### Closing issues

Closes #4654

### Self Check before Merge


packages/antd/src/providers/notificationProvider/index.tsx
```diff
@@ -1,41 +1,47 @@
...
-                key,
-                description: message,
-                message: description ?? null,
-                type,
-            });
-        }
-    },
-    close: (key) => notification.destroy(key),
+export const notificationProvider = (): NotificationProvider => {
+    const { notification } = App.useApp();
```
aliemir, line 8: This usage forces users to use the `App` component from `antd` without a notice in any docs, tsdoc typings or in console. For my understanding, this is a breaking change for the users with existing projects. 

It might not be if `antd` didn't had empty default values for the app context 😅 Check out this source code: https://github.com/ant-design/ant-design/blob/master/components/app/context.ts#L22

`useApp` will not throw if it's not used inside the `App` component but it will return `notification: {}`. We should add a check for this and use the previous `notification` instance (imported one) as a fallback. This will make sure we have the `notification`.

Also, I think we can move the `App.useApp` to a variable outside of the `notificationProvider` like;

```tsx
import { App, notification as staticNotification } from "antd";

const { useApp } = App;

export const notificationProvider = () => {
  const { notification: notificationFromContext } = useApp();
  const notification = "open" in notificationFromContext ? notificationFromContext : staticNotification;
}
```

There can be better ways to write this condition, leaving the implementation to you 😅 

This way will avoid releasing a breaking change and make sure we're keeping it working for the current users.

aliemir (changes requested):
Thank you for the PR @MahirMahdi, just left a comment to the `notificationProvider` implementation. We also need a changeset for the changes to be released. Check out the [contributing docs](https://refine.dev/docs/contributing) for more info about that.

packages/antd/src/providers/notificationProvider/index.tsx
MahirMahdi: Thank you so much. I should have tested the other projects, sorry I forgot. 😅

Sure, we can go ahead with this approach using the static `notification` instance as a fallback. I'll make the necessary changes tomorrow.

Also, we could update the documentation and the `ColorModeContextProvider` by passing `App` as a child so that projects built with `antd` can access the theme context from now on. 

Screenshot 2023-09-15 211317

If you're okay with that, I'll create a separate PR after I'm done with this one. Let me know what you think about it.😀

aliemir: Hey @MahirMahdi, thank you for the changes, looks great! 🚀 Yeah, I think adding `App` to `ColorModeContextProvider` makes sense for apps created with `create-refine-app` 🤔 I think we should also mention about the notifications in https://refine.dev/docs/api-reference/antd/theming/ this document, can you add a small section about it here? 🙏 

BTW, `create-refine-app` uses the templates inside [`superplate-core-plugins`](https://github.com/pankod/superplate-core-plugins) if you want to update them accordingly 🙏

MahirMahdi: Hey @aliemir, anytime. 😀 

Would you like me to create a separate PR for the document and plugin changes or should I include them in this PR?

aliemir: Let's make the documentation changes included in this PR and we can move to a separate one for the superplate changes

MahirMahdi: Sure. On it.

MahirMahdi: Here you go @aliemir. Let me know if it requires any changes.
Screenshot 2023-09-19 082446

MahirMahdi: @aliemir sorry to bother you. If it's okay then I'll update the PR.

aliemir: Hey @MahirMahdi sorry for the late response. Can you also add highlights to the code for the `<AntdApp>` lines using `// highlight-next-line` or `// highlight-start` and `// highlight-end`? 🙏 

Just noticed we also have some failing tests due to notifications, can you check those? Might be a type issue coming from the changes I've suggested 😅 

After these, we're good to go 🚀

MahirMahdi: Hey @aliemir , it's totally fine. 

I've added `// highlight-start` and  `// highlight-end`. But there's some issue with the useApp() context, working on it

aliemir: Thank you! Just send a push to fix the test issue with the notification provider and also added couple more tests 😅

MahirMahdi: No, thank you for helping me out. I never worked with Jest before. So was having a hard time figuring out the issue. Thank you.😀

aliemir (approved):
Thank you for the effort @MahirMahdi! I'm requesting couple more reviews from the team to make sure we're heading the right way here. I'm having some concerns about if we should mark this as breaking or not due to converting `notificationProvider` export from an object to function 🤔

.changeset/small-melons-sing.md
```diff
@@ -0,0 +1,7 @@
+---
+"@refinedev/antd": patch
+---
+
+fix: antd notificationProvider issue
+
+Antd notification component could not access theme context, now it's fixed.
```
aliemir, line 7:
````suggestion
Antd notification component could not access theme context, now it's fixed.

This release changes exported `notificationProvider` value from type `NotificationProvider` to `() => NotificationProvider`. If you previously had customizations applied to the `notificationProvider` object, you may need to update your code like the following:

```diff
- import { notificationProvider } from "@refinedev/antd";
+ import { notificationProvider as useDefaultNotificationProvider } from "@refinedev/antd";

- const myNotificationProvider = {
-    ...notificationProvider,
-    open: (...args) => {
-        // do some operation here
-        notificationProvider.open(...args);
-    },
- }
+ const myNotificationProvider = () => {
+     const notificationProvider = useDefaultNotificationProvider();
+     return {
+          ...notificationProvider,
+          open: (...args) => {
+             // do some operation here
+             notificationProvider.open(...args);
+          },
+     }
+ }
}

const App = () => {
    return (
        <Refine
            /* ... */
            notificationProvider={myNotificationProvider}
        >
            /* ... */
        </Refine>
    );
}
```
````

**Comment by MahirMahdi** [comment]
It's absolutely my pleasure @aliemir.

Absolutely, having different viewpoints will make it resistant to further issues. 

Hopefully, there won't be any issues as the 'mui' 'notificationProvider' is also exported as a function.

**Comment by MahirMahdi** [comment]
Hi @aliemir,no worries. I can take care of it and if I need help, I'll text you on Discord.😃

File: documentation/docs/api-reference/antd/theming.md [hunk 3]
```diff
@@ -162,6 +162,36 @@ const App: React.FC = () => {
 };
 ```
 
+### `useNotificationProvider` compatible with Theme
+
+The `notificationProvider` has been deprecated because it cannot consume the current `theme` context by default. To enable the `Notification` component to consume the current `theme` context, import the `App` package from `antd` and the `useNotificationProvider` package from `@refinedev/antd` and pass it as highlighted below:
```
aliemir, line 167:
```suggestion
The `notificationProvider` export has been deprecated because it cannot consume the current `theme` context by default. To enable the `Notification` component to consume the current `theme` context, import the `App` component from `antd` and the `useNotificationProvider` export from `@refinedev/antd` and pass it as highlighted below:
```

aliemir (approved):
🎖️

omeraplak (approved):

.changeset/small-melons-sing.md
```diff
@@ -0,0 +1,44 @@
...
+
+This release changes exported `notificationProvider` value from type `NotificationProvider` to `() => NotificationProvider`. If you previously had customizations applied to the `notificationProvider` object, you may need to update your code like the following:
+
+```diff
+- import { notificationProvider } from "@refinedev/antd";
++ import { notificationProvider as useDefaultNotificationProvider } from "@refinedev/antd";
```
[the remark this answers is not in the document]
MahirMahdi: I believe it would be `useNotificationProvider`, not `useDefaultNotificationProvider`.

.changeset/small-melons-sing.md
```diff
@@ -0,0 +1,44 @@
+---
+"@refinedev/antd": patch
+---
+
+fix: antd notificationProvider issue
+
+Antd notification component could not access theme context, now it's fixed.
+
+This release changes exported `notificationProvider` value from type `NotificationProvider` to `() => NotificationProvider`. If you previously had customizations applied to the `notificationProvider` object, you may need to update your code like the following:
```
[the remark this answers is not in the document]
MahirMahdi: Updated it.

MahirMahdi:
> I know it's been a long run 😄 But this is the final change, then we are good to go!
> 
> Thanks for the great contribution and follow-ups 🎖️

No problem at all. I'm glad I could help. 😃 

I was able to connect with amazing people like you, @omeraplak, and @aliemir because of this. So it's totally worth the wait. 🙌

BatuhanW (approved):

PR #4974, opened 2023-09-20
fix: add missing DataGrid operator conversion cases

### LINKED ISSUE #4970: [BUG] Some MUI Data Grid filter is not working with Hasura Provider
### Describe the bug

When using MUI's DataGrid and Hasura Provider and selecting 'starts with', 'ends with' or 'is any of' in the DataGrid default filter, I get the 'Operator <SELECTED_OPERATOR_NAME> is not supported' error.

---

Here is my guess for the situation.

The error seems to occur because the Hasura Provider throws an exception when an unsupported filter is specified in the Hasura Provider.
However, the Hasura Provider defines 'starts with' and 'ends with' as filters. 

Therefore, I suspect that this is not a bug specific to Hasura Provider, but a bug in useDataGrid, which provides props to the Data Grid. useDataGrid has a process for converting MUI operators to refine CRUDOperators. However, the conversion target for the three filters above is not implemented here. 

refs: https://github.com/refinedev/refine/blob/next/packages/mui/src/definitions/dataGrid/index.ts#L35-L69

The 'equals', 'isEmpty' and 'isNotEmpty' definitions, which are defined here, and 'contains', which can be referenced directly in the Hasura Provider, work correctly.

The title of this issue assumes that the MUI and Hasura Providers are used, but in reality other DataProviders may not be able to retrieve data correctly, just because they do not generate errors.

### Steps To Reproduce

1. Create MUI × Hasura playground with [refine.dev](https://refine.dev/#playground). (I chose Vite + MUI + Hasura + No Auth)
1. Filter record by one of the operators "starts with", "ends with", "is any of".
1. After a while, an error "Operator <SELECTED_OPERATOR_NAME> is not supported" is  occurs.

### Expected behavior

- "starts with" behaves as a starts with filter for each Provider.
- "ends with" behaves as an ends with filter for each Provider.
- "is any of" shall behave as an in filter for each Provider.

### Screenshot

https://github.com/refinedev/refine/assets/61409641/93bbe52d-f164-40ac-82ce-c75d8b390096

### Desktop

_No response_

### Mobile

_No response_

### Additional Context

There is also a bug that the "is not empty" filter does not work correctly if "namingConvention: "graphql-default" is specified in the Hasura provider.
However this is a bug on the Hasura provider side, so it has been fixed in another issue. I would like to submit it as a separate issue (This is because `_is_null` cannot be converted correctly as camelCase and the correct query cannot be issued).

IkumaTadokoro:

MUI defines the operator for each column types in [here](https://github.com/mui/mui-x/tree/2d09dbc6e5d03c4e66765d225ef93d3984e300fc/packages/grid/x-data-grid/src/colDef). However, there were not enough conversion cases for the following operators, so I added them to the mapping.

- isAnyof: used in Numeric, SingleSelect, String
- contains: used in String,
- startsWith: used in String
- endsWith: used in String

### Test plan (required)

I updated the tests with the new switch cases I've added.

The test case that was supposed to test the target was actually testing something else, so the describe part of that part has been changed in this PR.

### Closing issues

closes #4970

### Self Check before Merge

### DIFF
--- a/packages/mui/src/definitions/dataGrid/index.spec.ts
+++ b/packages/mui/src/definitions/dataGrid/index.spec.ts
@@ -201,7 +201,7 @@ describe("transformMuiOperatorToCrudOperators", () => {
     });
 });
 
-describe("transformMuiOperatorToCrudOperator", () => {
+describe("transformCrudOperatorToMuiOperator", () => {
     it("transform crud operator to mui operator with 'number' value", () => {
         expect(transformCrudOperatorToMuiOperator("eq", "number")).toEqual("=");
         expect(transformCrudOperatorToMuiOperator("ne", "number")).toEqual(
--- a/packages/mui/src/definitions/dataGrid/index.spec.ts
+++ b/packages/mui/src/definitions/dataGrid/index.spec.ts
@@ -184,6 +184,7 @@ describe("transformMuiOperatorToCrudOperators", () => {
         expect(transformMuiOperatorToCrudOperator("=")).toEqual("eq");
         expect(transformMuiOperatorToCrudOperator("!=")).toEqual("ne");
         expect(transformMuiOperatorToCrudOperator("not")).toEqual("ne");
+        expect(transformMuiOperatorToCrudOperator("isAnyOf")).toEqual("in");
         expect(transformMuiOperatorToCrudOperator("<")).toEqual("lt");
         expect(transformMuiOperatorToCrudOperator("before")).toEqual("lt");
         expect(transformMuiOperatorToCrudOperator("<=")).toEqual("lte");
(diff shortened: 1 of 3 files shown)


packages/mui/src/definitions/dataGrid/index.spec.ts
```diff
@@ -201,7 +201,7 @@ describe("transformMuiOperatorToCrudOperators", () => {
     });
 });
 
-describe("transformMuiOperatorToCrudOperator", () => {
+describe("transformCrudOperatorToMuiOperator", () => {
     it("transform crud operator to mui operator with 'number' value", () => {
         expect(transformCrudOperatorToMuiOperator("eq", "number")).toEqual("=");
         expect(transformCrudOperatorToMuiOperator("ne", "number")).toEqual(
```
BatuhanW, line 207: Could you update the tests with the new switch cases you've added?

BatuhanW (changes requested):
Hi @IkumaTadokoro, thanks for the contribution! Added a minor comment, then we can release it!

omeraplak (approved):

packages/mui/src/definitions/dataGrid/index.spec.ts
IkumaTadokoro: Tests have been added with [97ee1d0](https://github.com/refinedev/refine/pull/4974/commits/97ee1d03dbd652a80dd89bd0631faca1067cbace)! (I had overlooked that there was a test case for transformMuiOpeartorToCrudOperator. Thanks for that.)

I  also updated the test plan in the PR description.

BatuhanW (approved):

PR #5034, opened 2023-09-29
feat(core): add resource parameter to `useGo` hook

alicanerdurmaz:
Feature: The `go` function returned by `useGo` now accepts a `resource` object as the `to` parameter.
From now now, you can provide either a string or a resource object to the `go` function. When a resource object is passed, it will be transformed into the path defined within the resources array of the `<Refine />` component.

```tsx
import { useGo } from "@refinedev/core";

const MyComponent = () => {
    const go = useGo();

    return (
        <Button
            onClick={() => {
                go({
                    to:  {
                        resource: "posts", // resource name or identifier
                        action: "edit",
                        id: "1",
                    }
                    query: {
                         foo: "bar",
                    },
                    type: "push",
                });
            }}
        >
            Go Posts With Default Filters
        </Button>
    );
};
```

### Test plan (required)

image

### Self Check before Merge


aliemir (approved):

packages/core/src/hooks/router/use-go/index.tsx
```diff
@@ -1,15 +1,105 @@
...
+    const go = useCallback(
+        (config: GoConfigWithResource | GoConfigBase) => {
+            if (typeof config.to !== "object") {
+                return goFromRouter({ ...config, to: config.to });
+            }
+
+            // when config.to is an object, it means that we want to go to a resource.
+            // so we need to find the path for the resource.
+            const { resource } = resourceSelect(config.to.resource);
+            const newTo = findToPathFromResource(config.to, resource);
```
aliemir, line 55: I think we can also use `useGetToPath` for this

BatuhanW (approved):
Agree with @aliemir on using `useGetToPath`, other than that LGTM 👍🏼

.changeset/wicked-wolves-hunt.md
```diff
@@ -0,0 +1,42 @@
+---
+"@refinedev/core": minor
+---
+
+Feature: The `go` function returned by `useGo` now accepts a `resource` object as the `to` parameter.
```
omeraplak, line 5:
```suggestion
feat: The `go` function returned by `useGo` now accepts a `resource` object as the `to` parameter.
```

omeraplak (approved):

PR #5036, opened 2023-09-29
fix(simple-rest): fix persistent headers issue in custom method

### LINKED ISSUE #5035: [BUG] useCustomMutation is persisting the header
### Describe the bug

If we pass an extra header in useCustomMutation for an API call using the config property. For all the subsequent calls, we are getting the same header in the API calls. 
For more info, please check this out: https://discord.com/channels/837692625737613362/1156918397905420288

### Steps To Reproduce

Pass on an extra header in useCustomMutation using the config property for an API call & hit the API.
Then, for some other API call, hit the API using useCustomMutation with no headers passed.
Now, in the other API call, the same header will be present which was passed in the previous API call. This will be happening for all the subsequent API calls until the page is refreshed.

### Expected behavior

The extra headers should be ONLY passed for the API call in which the headers are explicitly mentioned in the useCustomMutation hook. They shouldn't be passed for any further calls.

### Screenshot

_No response_

### Desktop

_No response_

### Mobile

_No response_

### Additional Context

_No response_

saarthak08:
### Closing issues

Closes #5035 

### Self Check before Merge

### DIFF
--- a/.changeset/lemon-kings-hear.md
+++ b/.changeset/lemon-kings-hear.md
@@ -1,0 +1,5 @@
+---
+"@refinedev/simple-rest": minor
+---
+
+fix: headers persisting in useCustomMutation hook
--- a/.changeset/lemon-kings-hear.md
+++ b/.changeset/lemon-kings-hear.md
@@ -1,5 +1,5 @@
 ---
-"@refinedev/simple-rest": minor
+"@refinedev/simple-rest": patch
 ---
 
 fix: headers persisting in useCustomMutation hook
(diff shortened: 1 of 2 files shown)


.changeset/lemon-kings-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@refinedev/simple-rest": minor
```
aliemir, line 2:
```suggestion
"@refinedev/simple-rest": patch
```

aliemir (approved):
Thank you for the PR @saarthak08! I left a small change request in the changeset, I think it will be better to make a patch release 😅

alicanerdurmaz (approved):

omeraplak (approved):

Issue #4972, opened 2023-09-20
[BUG] Hasura Provide's "null" filter is not working with naming-convention: graphql-default


IkumaTadokoro:
### Describe the bug

Hasura and Hasura Provider can use either "hasura-default" or "graphql-default" as the naming convention for GraphQL fields.
When using the latter, attempting to retrieve a list with filter `null` and `nnull` will result in a "Error: field '_is_null' not found in type: ..." and data retrieval will fail.

The mappings "null" and "nnull" are defined as "_is_null", but if naming-convention: graphql-default is specified, they must be converted to camelCase, i.e. "_isNull". 

refs: https://github.com/refinedev/refine/blob/next/packages/hasura/src/utils/generateFilters.ts#L40-L45

### Steps To Reproduce

1. Create a project with the following configuration values.

```
❯ npm create refine-app@latest
✔ Downloaded remote source successfully.
✔ Choose a project template · refine-vite
✔ What would you like to name your project?: · metal-goats-flash
✔ Choose your backend service to connect: · data-provider-hasura
✔ Do you want to use a UI Framework?: · mui
✔ Do you want to add example pages?: · no
✔ Do you need any Authentication logic?: · none
✔ Do you need i18n (Internationalization) support?: · no
✔ Choose a package manager: · npm
```

2. Add namingConvention settings to Hasura Provider(src/App.tsx).

```tsx
  <Refine
-    dataProvider={dataProvider(client)}
+    dataProvider={dataProvider(client, { namingConvention: "graphql-default" })} 
```

3. Create a Hasura project for testing and set naming-convention to graphql-default according to the [documentation](https://hasura.io/docs/latest/schema/postgres/naming-convention/).
4. Add an appropriate resource for the list and generate the list display page with inferencer.
5. Add "null" or "nnull" filters to refine's data hooks（e.g. useList, useOne, useTable,...etc)

```ts
    filters: [
      {
        field: 'fieldName',
        operator: 'null',
        value: true,
      },
    ],
```

6. The above-mentioned bug occurs during data acquisition.

### Expected behavior

- null" is incorporated into the query as "_isNull: true".
- nnull" is incorporated into the query as "_isNull: false".

### Screenshot

![Visual Studio Code at 2023-09-20-17 14 10](https://github.com/refinedev/refine/assets/61409641/d200914e-c170-4c18-80de-bcb5bdeea4d0)

BatuhanW:
[...]

PR #5040, opened 2023-10-02
fix: convert hasura operator with graphql-default

IkumaTadokoro:

When using the Hasura provider with the 'graphql-default' naming convention, snake-case hasura operators (e.g. _is_null) were not converted to the correct case, causing query errors.
This problem is now fixed. Now the snake_case operator is converted according to the given naming convention (e.g. hasura-default: _is_null, graphql-default: _isNull).

### Test plan (required)

- Snake case hasura operator correctly converted to case added.
- The signature of the `generateFilters` function has changed, so existing test cases for this function will now run against the appropriate naming convention (but are backwards compatible in behaviour, as they are optional arguments).

### Closing issues

closes: #4972

### Self Check before Merge

### DIFF
--- a/packages/hasura/src/utils/generateFilters.ts
+++ b/packages/hasura/src/utils/generateFilters.ts
@@ -91,7 +91,7 @@ const convertHasuraOperatorToGraphqlDefaultNaming = (
 
 export const generateNestedFilterQuery = (
     filter: CrudFilter,
-    namingConvention: NamingConvention,
+    namingConvention: NamingConvention = "hasura-default",
 ): any => {
     const { operator } = filter;
 
@@ -125,8 +125,8 @@ export const generateNestedFilterQuery = (
 };
 
 export const generateFilters: any = (
-    namingConvention: NamingConvention,
     filters?: CrudFilters,
+    namingConvention: NamingConvention = "hasura-default",
 ) => {
     if (!filters) {
         return undefined;
--- a/packages/hasura/src/utils/generateFilters.ts
+++ b/packages/hasura/src/utils/generateFilters.ts
@@ -1,5 +1,7 @@
 import { CrudFilter, CrudFilters, CrudOperators } from "@refinedev/core";
+import camelcase from "camelcase";
 import setWith from "lodash/setWith";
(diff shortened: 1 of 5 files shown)


packages/hasura/src/utils/generateFilters.ts
```diff
@@ -102,19 +118,27 @@ export const generateNestedFilterQuery = (filter: CrudFilter): any => {
...
     return {
-        [`_${operator}`]: filter.value.map((f) => generateNestedFilterQuery(f)),
+        [`_${operator}`]: filter.value.map((f) =>
+            generateNestedFilterQuery(f, namingConvention),
+        ),
     };
 };
 
-export const generateFilters: any = (filters?: CrudFilters) => {
+export const generateFilters: any = (
```
aliemir, line 127: Shouldn't we also call the `generateFilters` with `namingConvention` in the data provider? @IkumaTadokoro

aliemir (changes requested):
Let's a tiny comment 😅 Thank you for the PR 🚀

packages/hasura/src/utils/generateFilters.ts
IkumaTadokoro: @aliemir 
Indeed, since the util is clearly dependent on the data provider, the format to pass it from there is better. Thanks for your review.

During the modification process I realised that currently the hasura live provider, unlike the data provider, does not support the naming convention.
Since it also references generateFilters, it is necessary to pass some naming convention, but the change to add naming convention to live provider is different from this PR scope, so I have modified it to pass `hasura-default` as a fixed convention once. 

If you would like me to make changes to this PR for that scope, please comment.

aliemir: Just send couple of commits to your branch. I thought the order of the parameters matter because there may be users who customized their hasura data providers and rely on these methods, we don't want to introduce a breaking change out of nowhere to them 😅 

Thank you for the quick work and thank you for your effort 🙏

aliemir (approved):

alicanerdurmaz (approved):

omeraplak (approved):

PR #5055, opened 2023-10-04
feat(cli): check core package is up to date for devtools

alicanerdurmaz:
feat: From now on, `npm run refine devtools init` updates `@refinedev/core` to latest version.

added 3 packages, renoved 1 package, changed 1 package, and audited 362 packages in 7s

Pasted Graphic 1

### DIFF
--- a/packages/cli/src/commands/devtools/index.ts
+++ b/packages/cli/src/commands/devtools/index.ts
@@ -75,13 +75,13 @@ const devtoolsInstaller = async () => {
     }
 
     console.log("🌱 Installing refine devtools...");
-    const installPackages = ["@refinedev/devtools@latest"];
+    const packagesToInstall = ["@refinedev/devtools@latest"];
     // we should update core package if it is lower than minRefineCoreVersionForDevtools
     if (isCorePackageOld({ pckg: corePackage })) {
-        installPackages.push("@refinedev/core@latest");
+        packagesToInstall.push("@refinedev/core@latest");
         console.log(`🌱 refine core package is updating for devtools...`);
     }
-    await installPackagesSync(installPackages);
+    await installPackagesSync(packagesToInstall);
 
     // empty line
     console.log("");
@@ -93,7 +93,7 @@ const devtoolsInstaller = async () => {
         "✅ refine devtools package and components added to your project",
     );
     // if core package is updated, we should show the updated version
-    if (installPackages.includes("@refinedev/core@latest")) {
+    if (packagesToInstall.includes("@refinedev/core@latest")) {
         const corePackageUpdated = await getRefineCorePackage();
         if (!corePackageUpdated) {
             throw new Error("refine core package not found");
--- a/packages/cli/src/commands/devtools/index.ts
+++ b/packages/cli/src/commands/devtools/index.ts
@@ -61,16 +61,16 @@ const devtoolsInstaller = async () => {
         throw new Error("refine core package not found");
     }
 
-    if (await isCorePackageDeprecated({ pckg: corePackage })) {
-        return;
-    }
-
-    const isInstalledDevtools = await spinner(
+    const isInstalled = await spinner(
         isDevtoolsInstalled,
         "Checking if devtools is installed...",
     );
-    if (isInstalledDevtools) {
+    if (isInstalled) {
         console.log("🎉 refine devtools is already installed");
+        return;
+    }
+
+    if (await isCorePackageDeprecated({ pckg: corePackage })) {
         return;
     }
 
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/devtools/index.ts
```diff
@@ -50,18 +56,32 @@ export const action = async (command: DevtoolsCommand) => {
...
         isDevtoolsInstalled,
         "Checking if devtools is installed...",
     );
-
     if (isInstalled) {
         console.log("🎉 refine devtools is already installed");
         return;
     }
 
+    if (await isCorePackageDeprecated({ pckg: corePackage })) {
```
BatuhanW, line 73: This gives me an impressions that it's simply returns boolean, and we aren't doing anything on false condition, but when navigate to this function, I see it handles messages and all. Maybe we can name it differently.

packages/cli/src/commands/devtools/index.ts
```diff
@@ -50,18 +56,32 @@ export const action = async (command: DevtoolsCommand) => {
...
+        return;
+    }
+
     console.log("🌱 Installing refine devtools...");
-    await installPackagesSync(["@refinedev/devtools@latest"]);
+    const packagesToInstall = ["@refinedev/devtools@latest"];
+    // we should update core package if it is lower than minRefineCoreVersionForDevtools
+    if (isCorePackageOld({ pckg: corePackage })) {
```
BatuhanW, line 80:
```suggestion
    if (shouldUpdateCorePackage({ pckg: corePackage })) {
```

packages/cli/src/commands/devtools/index.ts
```diff
@@ -72,6 +92,24 @@ const devtoolsInstaller = async () => {
     console.log(
         "✅ refine devtools package and components added to your project",
     );
+    // if core package is updated, we should show the updated version
+    if (packagesToInstall.includes("@refinedev/core@latest")) {
+        const corePackageUpdated = await getRefineCorePackage();
```
BatuhanW, line 97: `corePackageUpdated` should be a boolean. Is this package updated or not? Otherwise, we should use `updatedCorePackage`.

packages/cli/src/utils/refine/index.ts
```diff
@@ -20,3 +20,14 @@ export const hasDefaultScript = (): ReturnType => {
         dev: isDefault,
     };
 };
+
+export const getRefineCorePackage = async () => {
```
BatuhanW, line 24: Are we going to keep adding functions here as we need to get more packages? Maybe we can have a more generic use-case function for this one. Like `getPackage` which can accept string or array of strings.

packages/cli/src/commands/devtools/index.ts
```diff
@@ -124,8 +162,58 @@ const devtoolsInstaller = async () => {
...
+
     server();
 };
 
+export const isCorePackageOld = ({
+    pckg,
+}: {
+    pckg: { name: string; version: string };
+}) => {
+    return semver.lt(pckg.version, minRefineCoreVersionForDevtools);
```
BatuhanW, line 194: I don't think this is worth being a different function.

packages/cli/src/commands/devtools/index.ts
```diff
@@ -124,8 +162,58 @@ const devtoolsInstaller = async () => {
...
+        return;
+    }
+
     server();
 };
 
+export const isCorePackageOld = ({
+    pckg,
+}: {
+    pckg: { name: string; version: string };
```
aliemir, line 192: Since the `package` is a reserved keyword, maybe we can use `pkg` or `packageInfo` or similar instead of `pckg` 😅

packages/cli/src/utils/package/index.ts
```diff
@@ -42,7 +42,9 @@ export const getInstalledRefinePackages = async () => {
 
         const dependencies = JSON.parse(execution.stdout)?.dependencies || {};
         const refineDependencies = Object.keys(dependencies).filter(
-            (dependency) => dependency.startsWith("@refinedev"),
+            (dependency) =>
+                dependency.startsWith("@refinedev") ||
+                dependency.startsWith("@pankod"),
```
BatuhanW, line 47: Could this one cause a side-effect? Also we could use like `@pankod/refine-`.

BatuhanW (approved):
Added a minor comment, other than that, LGTM 🚀

packages/cli/src/utils/package/index.ts
alicanerdurmaz: Yes, it should be `@pankod/refine-`.

I don't think it will cause side effects. Because we are only using this function on `whoami`, `check-updates`, and `telemetry`.
Actually it's good to add.

aliemir (approved):

omeraplak (approved):

PR #5056, opened 2023-10-04
fix(devtools): conflicting `port` variable when using with `dev` command

aliemir:
Devtools server was picking up the wrong port to run when there's a `PORT` environment variable. This causes showing the wrong page content in the devtools panel.

### Self Check before Merge

### DIFF
--- a/.changeset/thin-dryers-hear.md
+++ b/.changeset/thin-dryers-hear.md
@@ -2,4 +2,4 @@
 "@refinedev/devtools": patch
 ---
 
-Use the proper devtools url coming from the websocket handshare rather than using the hardcoded port.
+Use the proper devtools url coming from the websocket handshake rather than using the hardcoded port.
--- a/.changeset/thin-dryers-hear.md
+++ b/.changeset/thin-dryers-hear.md
@@ -1,0 +1,5 @@
+---
+"@refinedev/devtools": patch
+---
+
+Use the proper devtools url coming from the websocket handshare rather than using the hardcoded port.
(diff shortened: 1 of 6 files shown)


alicanerdurmaz (approved):

.changeset/thin-dryers-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@refinedev/devtools": patch
+---
+
+Use the proper devtools url coming from the websocket handshare rather than using the hardcoded port.
```
BatuhanW, line 5:
```suggestion
Use the proper devtools url coming from the websocket handshake rather than using the hardcoded port.
```

packages/devtools-server/src/constants.ts
```diff
@@ -1,8 +1,8 @@
 export const DEFAULT_WS_PORT = 5002;
 export const DEFAULT_SERVER_PORT = 5001;
 
-export const WS_PORT = Number(process.env.WS_PORT || DEFAULT_WS_PORT);
-export const SERVER_PORT = Number(process.env.PORT || DEFAULT_SERVER_PORT);
+export const WS_PORT = DEFAULT_WS_PORT;
+export const SERVER_PORT = DEFAULT_SERVER_PORT;
```
BatuhanW, line 5: Should we have an option like `DEVTOOLS_PORT`?

aliemir: This PR is aimed to fix the issue with the `PORT` variable conflicts. The custom port feature is not hard to implement on the devtools side but will have issues with the authentication server when custom port is used.

We'll implement the feature in the future.

BatuhanW (approved):

omeraplak (approved):

PR #5160, opened 2023-10-24
fix: convert type of an option's value to string

### LINKED ISSUE #5157: [BUG] Mantine useSelect hook return value as type number instead of type string
### Describe the bug

For mantine Select, it only accepts value as string but the current useSelect hook return value type of number instead. It's kinda inconvenient to convert data type once again.

Current code of core useSelect doesn't actually convert the type of value:
Screen Shot 2023-10-24 at 12 47 38

Fix proposal:
a.toString() or String(a)

### Steps To Reproduce

 I reproduced the code: 
Screen Shot 2023-10-24 at 12 45 28
Screen Shot 2023-10-24 at 12 46 53

### Expected behavior

Each option should have `value` of type string

### Screenshot

_No response_

### Desktop

_No response_

### Mobile

_No response_

### Additional Context

_No response_

an-tran511:
Currently, option value is not converted into string after getting result from `useSelect` hook.
I refactor to use String() constructor for type conversion instead of type casting

### Test plan (required)

### Closing issues

Put `closes #5157` in your comment to auto-close the issue that your PR fixes (if such).

### Self Check before Merge


.changeset/quick-knives-jump.md
```diff
@@ -0,0 +1,5 @@
+---
+"@refinedev/core": major
```
omeraplak, line 2:
```suggestion
"@refinedev/core": patch
```

omeraplak (changes requested):

.changeset/quick-knives-jump.md
omeraplak: Thank you so much for your attention and effort! Can you mark the change as `patch` instead of `major`?

omeraplak (approved):

BatuhanW (commented):
Hello @an-tran511 

Thanks for the contribution! Looks good to me.

Before we merge it, could you add some test cases to cover your fix [in this file?](https://github.com/refinedev/refine/blob/90da1f82bba95edcf09a46a7fb51b760e4c711f2/packages/core/src/hooks/useSelect/index.spec.ts)

alicanerdurmaz (approved):

an-tran511:
> Hello @an-tran511
> 
> Thanks for the contribution! Looks good to me.
> 
> Before we merge it, could you add some test cases to cover your fix [in this file?](https://github.com/refinedev/refine/blob/90da1f82bba95edcf09a46a7fb51b760e4c711f2/packages/core/src/hooks/useSelect/index.spec.ts)

I added test case for this patch

BatuhanW (approved):

omeraplak (approved):

.changeset/quick-knives-jump.md
```diff
@@ -0,0 +1,5 @@
+---
+"@refinedev/core": patch
+---
+
+convert type of an option's value to string
```
aliemir, line 5: fix: convert type of an option's `value` to `string`

PR #5199, opened 2023-10-31
fix(use-select): correct types on `useSelect` options

### LINKED ISSUE #5192: [BUG] after upgrade to 4.44.9 optionLabel in useSelect not working
### Describe the bug

After upgrade to 4.44.9 the optionLabel in useSelect always return the default value which is title and after change it to another value it refused and it always return title and when title is not available is return the value which is the ID 

 

### Steps To Reproduce

Upgrade the refine core to 4.44.9 
using userSelect 
add option optionLabel 
```
const { selectProps: accomodationSelectProps } = useSelect({
    resource: "accomodation",
    optionLabel: "name",
  });
```

### Expected behavior

Returrn optionLabel: "name", as per the endpoint 

### Screenshot

_No response_

### Desktop

_No response_

### Mobile

_No response_

### Additional Context

_No response_

aliemir:
Exported `BaseOption` and `DefaultOption` types as successors of the deprecated `Option` type. `BaseOption` is `{ label: any; value: any; }` and `DefaultOption` is `{ label: string; value: any; }`.

Usage of the deprecated `Option` type was correctly assuming that the `value` property of the option is of type `string`. This assumption was wrong and now the types are changed to reflect the correct values of options with the ability to change it via 4th generic type `TOption` of `useSelect` hook.

---

Reverted the faulty assumption on option values of `useSelect` hook to be of type `string`. Now changed the types and the logic to reflect the correct values of options with the ability to change it via 4th generic type `TOption` of `useSelect` hook. (Reverted PR #5160)

By default `TOption` will be equal to `DefaultOption` type which is `{ label: string; value: any; }`. If you want to change the type of options, you can do it like this:

```tsx
import { useSelect, HttpError } from "@refinedev/core";

type MyData = {
    id: number;
    title: string;
    description: string;
    category: { id: string };
};

type Option = { label: MyData["title"]; value: MyData["id"] }; // equals to { label: string; value: number; }

useSelect<MyData, HttpError, MyData, Option>({
    resource: "posts",
});
```

### Closing issues

Fixes #5192

### Self Check before Merge

### DIFF
--- a/packages/core/src/hooks/useSelect/index.ts
+++ b/packages/core/src/hooks/useSelect/index.ts
@@ -7,7 +7,8 @@ import get from "lodash/get";
 import { useList, useMany, useMeta } from "@hooks";
 import {
     CrudSorting,
-    Option,
+    DefaultOption,
+    BaseOption,
     BaseRecord,
     GetManyResponse,
     GetListResponse,
@@ -38,16 +39,13 @@ export type UseSelectProps<TQueryFnData, TError, TData> = {
      * Set the option's value
      * @default `"title"`
      */
-    optionLabel?: keyof TData extends string
-        ? keyof TData
-        : never;
+    optionLabel?: keyof TData extends string ? keyof TData : never;
+    // optionLabel?: String | keyof TData;
     /**
      * Set the option's label value
      * @default `"id"`
      */
-    optionValue?: keyof TData extends string
-        ? keyof TData
-        : never;
+    optionValue?: keyof TData extends string ? keyof TData : never;
     /**
      * Allow us to sort the options
      * @deprecated Use `sorters` instead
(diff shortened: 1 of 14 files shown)


omeraplak (approved):

packages/core/src/hooks/useSelect/index.ts
```diff
@@ -38,16 +38,13 @@ export type UseSelectProps<TQueryFnData, TError, TData> = {
      * Set the option's value
      * @default `"title"`
      */
-    optionLabel?: keyof TData extends string
-        ? keyof TData
-        : never;
+    optionLabel?: keyof TData extends string ? keyof TData : never;
+    // optionLabel?: String | keyof TData;
```
alicanerdurmaz, line 42:
```suggestion
   
```

alicanerdurmaz (approved):

Issue #5317, opened 2023-11-26
[BUG] - CSV Export uses an outdated pacakge - export-to-csv-fix-source-map


vikavorkin:
### Describe the bug

Current version of the `export-to-csv` package is `export-to-csv-fix-source-map@0.2.1` 

This version has a bug where not all headers are exported and it depends on the first data item.

This issue affect `@refinedev/core` since `useExport` was added @ [8a115b5f](https://github.com/refinedev/refine/commit/8a115b5f6bcfaa8db6126dfff3adb6d0890923ce) 

### Steps To Reproduce

1. Load a table with the following data 
``` json
[
 {"field1":"foo"}, 
 {"field1":"bar", "field2": "baz"}
]
```
2. Export a CSV using
``` ts
    const { triggerExport, isLoading } = useExport<IVendor>({
        mapData: item => flatten(item),
        exportOptions: {
            showLabels: true,
            // There is an issue if the first item in the list is missing a keys, no column will be generated for them.
            // Manually defining the keys
            useKeysAsHeaders: false,
            headers: ["field1", "field2"]
        }
    });
```
4. CSV will not contain "field2"

### Expected behavior

Specifying the headers for the CSV should export all the headers without depending on the actual data for the keys.

aliemir:
Hey @vikavorkin, sorry for the issue! `export-to-csv-fix-source-map` was a fix for webpack back then, now it's most likely not needed. We can switch to `export-to-csv` and fix this issue. We'll be happy to accept your contribution if you want to open up a PR for this 🙏

vikavorkin:
Hi @aliemir, I will gladly open a PR, will take me some time to get to it.
The current parameters look to match the latest version so I expect no breaking changes.

Will update, thanks for the quick answer! 👍

omeraplak:
> Hi @aliemir, I will gladly open a PR, will take me some time to get to it. The current parameters look to match the latest version so I expect no breaking changes.
> 
> Will update, thanks for the quick answer! 👍

Thank you, I have assigned the issue to you. You can take a look at our Contributing document at [https://refine.dev/docs/contributing/]

PR #5332, opened 2023-11-29
fix: replace export-to-csv-fix-source-map with papaparse

vikavorkin:
This fixes https://github.com/refinedev/refine/issues/5317, the current package used to generate the CSV is old and has few issues, this PR changes the implentation to use the same package as for `useImport` - `papaparse` for the CSV generation.

### Test plan (required)

Due to changes regarding the underlying options (from options for `export-to-csv` to the options of `papaparse`)
* Add tests to check backward compatibility of the old options as not to break interface.

Closing issues

closes https://github.com/refinedev/refine/issues/5317

### Self Check before Merge

-   [X] Corresponding issues are created/updated or not needed
-   [X] Docs are updated/provided or not needed
-   [X] Examples are updated/provided or not needed
-   [X] TypeScript definitions are updated/provided or not needed
-   [X] Tests are updated/provided or not needed
-   [X] Changesets are provided or not needed


vikavorkin:
I did find a difference in the options the underlying library uses that may break code that uses those option.

From what I see, this fix can't be a patch as-is, do you want me to add backward-compatibility support (mapping the old options to the new ones) or release it as a minor/major change?

| Old              |      New                             |  
|---------------|:-------------------------------:|
| showLabels  |  showColumnHeaders        |
| headers        |    columnHeaders              |

vikavorkin:
Build does fail with the error
```
Module not found: ESM packages (export-to-csv) need to be imported. Use 'import' to reference the package instead. https://nextjs.org/docs/messages/import-esm-externals
```

From what I see it's related only to `live-previews` and `export-to-csv` been ESM but the build processes changes it to `require`.

Not sure how to fix this.

vikavorkin:
Adding 
```
    experimental:{
        esmExternals: 'loose'
    }
```
to the `packages/live-previews/next.config.js` fixes the issue but I don't think this is correct as it's an issue with `export-to-csv`.

aliemir:
Hey @vikavorkin thank you for the PR and the investigation about the issue 🙏 For now, we're not aiming to convert refine to an ESM only package 🤷 I'm not sure if we can succeed if we try to import ESM in CJS and make it compatible for everyone 😅 

Instead, we discussed an another option to resolve this issue. We already use `papaparse` package in `useImport` hook and it also provides an export option (`unparse` -what an unusual name 🤣-)

To avoid any breaking changes, we may keep the existing `export-to-csv-fix-source-map` for its types and provide backward compatible properties in `exportOptions` prop. 

This means we should handle some stuff in our codebase.

---

**Converting to `papaparse.unparse` from `export-to-csv.ExportToCsv`**

The configuration is a bit different but I think this still can be done easily considering the job of both functions are converting arrays to csv 😅 

**Downloading the CSV**

This was previously handled by the `export-to-csv` package which appends an anchor element with `download` attribute and triggers a click, then removes it from DOM. We can replicate that in our codebase to keep `useExport` making auto downloads. This can be done only for browser, so let's keep in mind to have a `typeof document` check before doing any actions.

We're pushing to make Refine work on any platform and not just browsers, so this might be a good chance to update the `triggerExport` command to return `Promise<string>` instead of `Promise<void>`. We can return the generated csv and for browsers still trigger a download. This will give ability to use `useExport` in other environments. (Nice to have but is not really in scope of this PR)

**Mapping Old Parameters**

I just checked the config types for the `export-to-csv-fix-source-map`:

```tsx
export interface Options {
    filename?: string;
    fieldSeparator?: string;
    quoteStrings?: string;
    decimalSeparator?: string;
    showLabels?: boolean;
    showTitle?: boolean;
    title?: string;
    useTextFile?: boolean;
    useBom?: boolean;
    headers?: string[];
    useKeysAsHeaders?: boolean;
}
``` 

Let's see if we can find a respective option in papaparse for each of them 😅 

`filename` is for download and after switching to `papaparse` will be handled by us.
`fieldSeperator` is `delimiter`
`quoteStrings` is `quotes`
`decimalSeparator` I think `papaparse` doesn't have this option (We can skip It I guess and redirect users to `mapData` in TSDoc)
`showLabels` this is removed in the latest version of `export-to-csv` but looks like its for the headers (showing or hiding them) and I think `papaparse` can control this via `header: false` 🤔 
`showTitle` and `title` `papaparse` doesn't have this and we can prepend it in our code.
`useTextFile` this is also related with the download. `papaparse.unparse` returns a string, the extension of the dowloaded file is up to us.
`useBom`: Bom is `\ufeff` added to the start of the file. We can handle this as well.
`headers`: Explained below but I think `columns` will cover it.
`useKeysAsHeaders`: this is same as `columns` being unset. 

I think that's all. If you have any other issues regarding the mapping of the parameters, let's get in touch! 

**Fixing the issue of this PR**

This was the main issue of this PR, the exported csv does not have the defined headers if data is not consistent for all items. We can use the `columns` property of `UnparseConfig` along with the `header` property. I think we can just map `headers` to `columns` which essentially works the same. (We may need to check the source or documentation to make sure this is correct but looks like it)

**Adding Tests**

This will be crucial for us to make sure we're making it backward compatible. Existing tests are expected to work same and we can add additional ones for new configuration etc.

---

There are the expected changes for this PR, let's discuss if you have any other ideas. Let us know if you can spare some time to implement this or we can add this to our to-do and let our team work on this 🚀  A long comment may look a bit scary to handle but I think the changes will not be much complicated and possible to do with ease.

vikavorkin:
I have changed `useExport` to use `papaparse` and `file-saver`.
Added few tests, need to add more - WIP.

vikavorkin:
Hi,
Added what you asked for, sorry I missed what you wrote about the return type of `triggerExport`.
I tried to avoid using the generic download a file via data url by using `file-saver` package. I had trouble supporting all the different browser implementations and `file-saver` should provide a much more solid solution.


Generally I would prefer `triggerExport` return the CSV string and have the user handle it.
For now I added your implementation for the download.

aliemir (approved):
Amazing work, congrats! Thank you for the PR, I just had the chance to review and test it out 🚀 🚀 

This PR is queued to be released with our January release and will be included in the release, thanks again for the great effort! 🙏

.changeset/late-insects-fly.md
```diff
@@ -0,0 +1,7 @@
+---
+"@refinedev/core": patch
+---
+
+fix: replace export-to-csv-fix-source-map with papaparse
+
+Replace useage of `export-to-csv-fix-source-map` to `papaparse`, fixing issues with `useExport` hook.
```
aliemir, line 7:
```suggestion
Replace usage of `export-to-csv-fix-source-map` to `papaparse`, fixing issues with `useExport` hook.
```

packages/core/src/hooks/export/index.ts
```diff
@@ -175,15 +200,66 @@ export const useExport = <
...
+            }
+        }
+
+        let csv = unparse(rawData.map(mapData as any), unparseConfig);
+        if (options.showTitle) {
+            csv = options.title + '\r\n\n' + csv
+        }
+
+        // Backward compatibility support for downloadInBrowser of the exported file, only works for browsers.
+        if(window && csv.length > 0){
```
aliemir, line 248: I think checking it as `if(window)` will throw a `ReferenceError` in environments that its not defined.

```suggestion
        if(typeof window !== "undefined" && csv.length > 0){
```

aliemir: I actually like the idea of having a prop like `download: boolean;` to let the user decide whether to download or not but I don't want to ask for a feature in a completed PR 😅 We can work on it after this is merged.

vikavorkin: Added the `download: boolean` option in separate commit.

aliemir:
Thank you for the update @vikavorkin 🙏 Can you update the documentation to reflect the current changes? Check out the docs section in the [Contributing](https://refine.dev/docs/guides-concepts/contributing/#working-on-documentation) guide

vikavorkin:
Hi @aliemir, sorry had some trouble launching the docs and not enough time to troubleshoot it.
I will try to have something this weekend.

.changeset/late-insects-fly.md
```diff
@@ -0,0 +1,7 @@
+---
+"@refinedev/core": patch
+---
+
+fix: replace export-to-csv-fix-source-map with papaparse
```
omeraplak, line 5:
```suggestion
fix: replace export-to-csv-fix-source-map with papaparse #5317
```

omeraplak (approved):

alicanerdurmaz:
@vikavorkin Hello again, thanks for the PR. It looks great.

I updated `useExport` documentation, I would be very happy if you could review it when you are available.

alicanerdurmaz (approved):

aliemir (approved):

PR #5382, opened 2023-12-17
chore: lock the project to node v18.17

neuodev:
- Lock the project to node `v18.17.0` to avoid having unnecessary diff in `package-lock.json`.  
- Add missing TS types for `is-git-clean` package.


package.json
```diff
@@ -45,6 +45,7 @@
     "@lerna/filter-options": "^6.2.0",
     "@netlify/plugin-nextjs": "4.32.2",
     "@nrwl/nx-cloud": "^15.0.2",
+    "@types/is-git-clean": "^1.1.2",
```
BatuhanW, line 48: Can we move this change to `packages/codemod/package.json`?

.nvmrc
```diff
@@ -0,0 +1 @@
+v18.17.0
```
BatuhanW, line 1:
```suggestion
```

Let's remove this change, as locking into a specific version is not beneficial. And also not everyone is using nvm.

BatuhanW (changes requested):
Hello @neuodev thanks for the PR. Added a review 🚀

PR #5386, opened 2023-12-18
[FEAT] Allow objects to be translated in i18nProvider

### LINKED ISSUE #5364: [FEAT] Allow objects to be translated in i18nProvider

In certain cases it makes sense to store translations as JSON objects (with locales as keys) in the database. It would be great if the types of the `i18nProvider` could be extended to allow objects as well as the first parameter of the `translate` function so we can easily translate them as well.

An alternative is of course a function in user land.

### Additional context

_No response_

I would like to see the following type in `@refinedev/core/src/contexts/translation/ITranslationContext.ts`:
````typescript
export type Translate = (
    key: string | Record<string, string>, // here is the change
    options?: any,
    defaultMessage?: string,
) => string;
````

jaulz:
This PR closes #5364 and is basically a simple type change in the definition of the `translate` function. The `translate` function previously only allowed keys to be passed as string which could then be used to retrieve the translation from the real i18n provider via a mapping table. The new interface also allows objects to be passed as keys which for example then allows you to pass an object with locale -> translation mapping:
````typescript
    const i18nProvider: I18nProvider = {
        translate: (key, params) => {
          if (typeof key === 'object') {
            return key[i18n.language] ?? '';
          }
          
          return t(key, params)
        },
        changeLocale: (lang) => i18n.changeLanguage(lang),
        getLocale: () => i18n.language,
    };
````

````typescript
const translate = useTranslate();
const title = {
  en: 'Switzerland',
  de: 'Schweiz',
}
  
console.log(
  "Translate",
  translate(title),
);
````

### Test plan (required)

Since the change is only a type change, the implementation itself is up to the end user and thus I think there is no real test needed. However,  I tried to adapt all the existing examples and even improve them by using the `I18nProvider` interface instead of fixed types.

### Closing issues

Closes #5364

### Self Check before Merge

-   [X] Corresponding issues are created/updated or not needed
-   [X] Docs are updated/provided or not needed
-   [X] Examples are updated/provided or not needed
-   [X] TypeScript definitions are updated/provided or not needed
-   [X] Tests are updated/provided or not needed
-   [X] Changesets are provided or not needed


aliemir:
I added a comment to the related issue of this PR here https://github.com/refinedev/refine/issues/5364#issuecomment-1862426688

We've discussed the outcome and decided that it will break the existing i18n provider implementations both on type level and if the first argument is treated as string then it will also break the runtime.

Closing this for now, thank you for the contribution #:pray:

PR #5389, opened 2023-12-18
fix(link-and-title) : Change link to tag path and change subtitle

SeungChans:

### Test plan (required)

### Closing issues

### Self Check before Merge

### DIFF
--- a/packages/antd/src/components/pageHeader/index.tsx
+++ b/packages/antd/src/components/pageHeader/index.tsx
@@ -31,7 +31,7 @@ export const PageHeader: FC<AntdPageHeaderProps> = ({ children, ...props }) => {
         );
 
     const subtitle =
-        typeof props.title === "string" ? (
+        typeof props.subtitle === "string" ? (
             <Typography.Title
                 className={RefinePageHeaderClassNames.SubTitle}
                 level={5}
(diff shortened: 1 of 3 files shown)


packages/antd/src/components/pageHeader/index.tsx
```diff
@@ -31,7 +31,7 @@ export const PageHeader: FC<AntdPageHeaderProps> = ({ children, ...props }) => {
         );
 
     const subtitle =
-        typeof props.title === "string" ? (
+        typeof props.subtitle === "string" ? (
```
aliemir, line 34: It should be `props.subTitle` according to the API of Ant design.

packages/antd/src/components/pages/auth/components/login/index.tsx
```diff
@@ -268,7 +268,7 @@ export const LoginPage: React.FC<LoginProps> = ({
                             "Don’t have an account?",
                         )}{" "}
                         <ActiveLink
-                            to="/register"
+                            to={registerLink}
```
aliemir, line 271: This prop doesn't work this way. `registerLink` is expected to be a `React.ReactNode` and had this check right before this code block. Same also applies to the change in the `register/index.tsx` about the `loginLink` prop. 

These changes are redundant and should be reverted.

aliemir (changes requested):
Hey @SeungChans, we cannot accept this PR right now. Due to lack of information about the issue and solution it provides. Please open up an issue first and link the issue to the PR. 

Also please provide a changeset, you can check out the [Contributing](https://refine.dev/docs/guides-concepts/contributing/) guide to get started.

Closing this for now. Please do not hesitate to re-open when you're done with the requests I mentioned above.

PR #5423, opened 2023-12-26
fix(core): include all keys when propagating errors to fields

aliemir:
Updated `flattenObjectKeys` method to support both nested and non-nested variables when propagating server side errors to form fields.

### Self Check before Merge


omeraplak (approved):

.changeset/heavy-shrimps-own.md
```diff
@@ -0,0 +1,5 @@
+---
+"@refinedev/core": patch
+---
+
+Updated `flattenObjectKeys` method to support both nested and non-nested variables when propagating server side errors to form fields.
```
aliemir, line 5:
```suggestion
Updated `flattenObjectKeys` method to support both nested and non-nested variables when propagating server side errors to form fields. Resolves [#5461](https://github.com/refinedev/refine/issues/5461)
```

PR #5457, opened 2023-12-29
feat(core): add ability to call custom methods from data providers

aliemir:
**Example Usage**

 ```tsx
 const { mutate } = useCustomMethod("myCustomMethod", { as: "mutation" });
 const { data, isLoading } = useCustomMethod("myCustomMethod", { as: "query" });
 const { data, isLoading } = useCustomMethod("myCustomMethod");
 ```

**Example Implementation**

 ```tsx
 import simpleRestDataProvider from "@refinedev/simple-rest";

 const dataProvider = {
   ...simpleRestDataProvider("https://my.api.com"),
   customMethods: {
     myCustomMethod: ({ params, meta }) => ({
       data: {},
     }),
   },
 };
 ```

### Self Check before Merge


.changeset/four-bottles-explain.md
```diff
@@ -0,0 +1,29 @@
...
+"@refinedev/core": minor
+---
+
+// TODO
+data provider custom methods
+
+**Example Usage**
+
+```tsx
+const { mutate } = useCustomMethod("myCustomMethod", { as: "mutation" });
```
BatuhanW, line 11: I think we should add 2 hooks:

- useCustomQuery
- useCustomMutation

And Data Provider with


```
const dataProvider = {
  ...,
  customMutations: { ... },
  customQueries: { ... },
}
```
---

- `useCustomMethod` itself isn't clear, if you don't have the context. `method` has many meanings in different contexts.

- Implementation, testing and maintenance also is harder with a single hook because of cyclomatic complexity.

- I think having 2 separate methods has better UX and ergonomics:

  - It's easier to mistake mutation with query. For example user may pass `as: "mutation"`, even though the method is `query`.

  - If user tries to use `useCustomMutation` hook, and if it's not in `customMutations` field, we can throw an error. With current implementation, AFAIK, we can't prevent this.

  - if I need to use 5 mutations, instead of just importing `useCustomMutation` hook, I'll import `useCustomMethod` and repetitively pass 2nd parameter `{ as: "mutation" }` to each of them.

BatuhanW (changes requested):

Issue #5472, opened 2024-01-04
[FEAT] Ability to configure `useCan`'s queryOptions globally and through `CanAccess` component.


BatuhanW:
### Is your feature request related to a problem? Please describe.

Currently, `useCan` hook already accepts `queryOptions` parameter. This is useful when users wants to modify `cacheTime` or `staleTime` options to improve performance.

However, currently it's not possible to modify queryOptions for `CanAccess` component, or UI components using `CanAccess` under the hood, such as `EditButton` of `@refinedev`'s `antd`, `mui`, `mantine`, `chakra-ui` packages.

### Describe alternatives you've considered

It's possible to swizzle these components and update their `useCan` calls, but that's irreversible and overkill, just to cache AC queries.

### Describe the thing to improve

1-) Global Configuration

Update `AccessControlProvider` interface to accept optional `queryOptions`.

```tsx
<Refine ....
  accessControlProvider={{
    can: ....
    options: {
      queryOptions: {  staleTime: 5000 //  UseQueryOptions<CanReturnType> }
    }  
  }}
```

2-) Component prop

Add `queryOptions` prop to `CanAccess` component.

```tsx
<CanAccess queryOptions={{ cacheTime: 25000 }} />
```

Checklist:
- Refactor `useCan` hook to respect `queryOptions` in the following prioritization order if provided: `hook parameter -> component prop -> global config`
- Add new specs.
- Update relevant documentation sections.

akkyyypatel:
The understanding of global configuration is good

Conqxeror:
Is someone working on this issue or should I take it? @BatuhanW @alicanerdurmaz @aliemir

BatuhanW:
Hey @Conqxeror, assigned to you. Looking forward to your PR!

Conqxeror:
If my code is right as it was described in issue (I might have misunderstood the issue, even got confused while solving) then I can go further with docs? @BatuhanW

BatuhanW:
> If my code is right as it was described in issue (I might have misunderstood the issue, even got confused while solving) then I can go further with docs? @BatuhanW

Your code looks good, added some minor comments for adding specs.

For documentation, you can update the following docs:
- https://github.com/refinedev/refine/blob/master/documentation/docs/authorization/access-control-provider/index.md
- https://github.com/refinedev/refine/blob/master/documentation/docs/authorization/components/can-access/index.md

PR #5483, opened 2024-01-05
feat: Ability to configure useCan's queryOptions globally and through CanAccess component

Nabeelcodes110:
Made the changes according to the issue  #5472.
please review the changes and let me know if it is correct or not.

### DIFF
--- a/packages/core/src/components/canAccess/index.tsx
+++ b/packages/core/src/components/canAccess/index.tsx
@@ -1,5 +1,9 @@
 import React, { useEffect } from "react";
-
+import {
+    useQuery,
+    UseQueryOptions,
+    UseQueryResult,
+} from "@tanstack/react-query";
 import { useCan, useResource } from "@hooks";
 import { BaseKey, IResourceItem, ITreeMenu } from "../../interfaces";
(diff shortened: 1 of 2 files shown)


packages/core/src/contexts/accessControl/IAccessControlContext.ts
```diff
@@ -28,6 +33,7 @@ export type CanReturnType = {
 export interface IAccessControlContext {
     can?: ({ resource, action, params }: CanParams) => Promise<CanReturnType>;
     options?: {
+        queryOptions?: UseQueryOptions<CanReturnType>;
```
BatuhanW, line 36: You are accepting these queryOptions but it's not used anywhere.

packages/core/src/components/canAccess/index.tsx
```diff
@@ -59,6 +64,7 @@ export const CanAccess: React.FC<CanAccessProps> = ({
     fallback,
     onUnauthorized,
     children,
+    queryOptions,
```
BatuhanW, line 67: Please add specs.

PR #5511, opened 2024-01-14
fix(core): useform onFinishUpdate promise settlement

### LINKED ISSUE #5460: [BUG] useForm onFinishUpdate promise never settles when in pessimistic mutationMode
### Describe the bug

The useForm hook is used with mutationMode set to "pessimistic" and the update mutation from the data provider rejects.
The issue in the code is here:  https://github.com/refinedev/refine/blob/master/packages/core/src/hooks/form/useForm.ts#L581
In that code path, the promise executor returns without ever calling reject (nor resolve) so the promise is never settled. This means the onFinish function never rejects (nor resolves) and it also means the underlying react-hook-form state remains with `isSumitting` as true indefinitely.

### Steps To Reproduce

1. Implement a component with useForm configured with mutationMode set to "pessimistic"
2. Add a then and catch to the onFinish method with console log statements to tell when its promise resolves or rejects
3. Add a console log to log the isSubmitting form state on every render of the component
4. Submit a form update mutation that fails (rejecting the promise)
5. Monitor the isSubmitting form state, noting that it remains true
6. Note the log statements from the onFinish resolve nor reject are ever called

### Expected behavior

The promise returned by onFinish should reject and the form state should revert to isSubmitting set to false.
A simple fix would be to just remove the "return" from line 581 so that `reject` is called after `onMutationError` executes. 
(Perhaps reject should be called with the error passed in as an argument, but that may be a bigger and undesired change.)

### Screenshot

_No response_

### Desktop

_No response_

### Mobile

_No response_

### Additional Context

_No response_

ShivamDureja:
### Issue
This fixes #5460

### Description
This PR addresses the issue where the `useForm` hook, specifically in the code path for pessimistic mutationMode, had a promise executor that returned without ever calling reject (nor resolve). This resulted in the promise not being settled, causing the `onFinishUpdate` promise never to reject (nor resolve). 

### Changes
The `return` statement from line 581 in the useForm hook has been removed to ensure that the promise executor now correctly calls reject after the `onMutationError` executes. This adjustment guarantees that the `onFinishUpdate` promise is now set to reject, and the form state is restored with `isSubmitting` set to `false`.

### Self Check before Merge


packages/core/src/hooks/form/useForm.ts
```diff
@@ -578,7 +578,7 @@ export const useForm = <
                         },
                         onError: (error: TResponseError, _, context) => {
                             if (onMutationError) {
-                                return onMutationError(error, values, context);
+                                onMutationError(error, values, context);
```
aliemir, line 581: @ShivamDureja can we add a case to our `useForm` tests to check for this? Let's make sure we're handling this in our tests to be safe for the future.

.changeset/tame-poems-deny.md
```diff
@@ -0,0 +1,13 @@
+---
+"@refinedev/core": minor
```
aliemir, line 2:
```suggestion
"@refinedev/core": patch
```

I'm suggesting this to be marked as `patch` rather than `minor` 🙏

.changeset/tame-poems-deny.md
```diff
@@ -0,0 +1,13 @@
...
+---
+
+---
+
+## "refinedev/core": minor
+
```
aliemir, line 8:
```suggestion
```

We don't need this heading here, changesets will handle this for us when creating GitHub releases 🙏

aliemir (changes requested):
Thank you for your effort @ShivamDureja 🙏 I've made some comments to your changes, shouldn't be hard to update but let me know if you have any issues 🚀

packages/core/src/hooks/form/useForm.ts
BatuhanW: @ShivamDureja you need to modify this file: https://github.com/refinedev/refine/blob/master/packages/core/src/hooks/form/useForm.spec.tsx

aliemir, line 581: This line is repeated for the `"create"` action and has the same problem.

aliemir (changes requested):
Thank you for the effort! I've noticed that the required changes have only made for the "update" action and not for the "create" action. Can you please update the necessary lines as well? (L462 on useForm.ts)

packages/core/src/hooks/form/useForm.ts
ShivamDureja: Thanks for the review @aliemir I have made changes accordingly

BatuhanW (changes requested):
Currently, tests are failing, since now we don't have early `return onMutationError` now, `reject()` is being called, causing tests to fail. I'm not sure if we need to return new Promise with setTimeout here at all, need to discuss. @aliemir @omeraplak

BatuhanW:
Hey @ShivamDureja thanks for the PR! We'll be refactoring this hook more throughly, as there are many other issues that needs to be fixed together. We'll handle these issues under issue https://github.com/refinedev/refine/issues/5702.

PR #5526, opened 2024-01-17
chore(core): mark `dataProvider` prop as optional

aliemir:
Updated Refine context and `<Refine />` component props to make `dataProvider` prop optional.

By default if there are no data providers are set and a data hook is invoked, it will throw an error.

Marking it as optional will make users get started more easily rather than having to set up a dummy data provider with a fake API or something.

### Test plan (required)

There were some tests that were using data hooks/methods but didn't properly set up a test wrapper with a data provider. Updated them to not throw any errors. 

Existing tests of `useDataProvider` is also testing it for throwing when data provider is not defined.

### Self Check before Merge


packages/core/src/components/containers/refine/index.tsx
```diff
@@ -294,7 +294,7 @@ export const Refine: React.FC<RefineProps> = ({
                         {...(authProvider ?? {})}
                         isProvided={Boolean(authProvider)}
                     >
-                        <DataContextProvider {...dataProvider}>
+                        <DataContextProvider {...(dataProvider ?? {})}>
```
BatuhanW, line 297: I think we don't need this fallback `?? {}`.

BatuhanW (approved):

packages/core/src/components/containers/refine/index.tsx
arghyapolley: I think <DataContextProvider {...dataProvider && {...dataProvider}}> this can serve better efficiency.

PR #5610, opened 2024-02-07
feat(core): added custom optionLabel

### LINKED ISSUE #4880: [FEAT] custom optionLabel and remove duplicate list in useSelect hook.

I wanted to add a custom optionLabel in useSelect hook which I couldn't manipulate in useSelect options, @omeraplak have helped me with creating a custom options which resolved my issues as below:
```
const { selectProps, queryResult, defaultValueQueryResult } = useSelect({
        resource: "resource",
});

const allOptions = [
    ...(queryResult?.data?.data || []),
    ...(defaultValueQueryResult?.data?.data || [])
];

// in select components I have created my custom option label as below.
<Select
    {...selectProps}
    options = {allOptions.map((p: IPoint) => ({
        label: `${p.location_name} - ${p.landmark}`,
        value: p.id
    }))}
/>
```
This was working fine but whenever I search in my select input box my default value get pushed to the list which result in a duplicate value showing multiple times.

I have tried to remove duplicate value by creating a function that filters out duplicate value as below:
```
const uniqueArrayofObject = <TData extends { id: number | string }>(options: TData[]): TData[] => {
    return Array.from(new Set(options.map(a => a.id)))
      .map(id => options.find(a => a.id === id)) as TData[];
}
```
And using this function on my allOptions:
```
const allOptions = uniqueArrayofObject<IPoint>([
    ...(queryResult?.data?.data || []),
    ...(defaultValueQueryResult?.data?.data || [])
]);
```

### Additional context

_No response_

If we could have an option to create a custom optionLabel and filter out data based on unique value from the record that could solve this issue. As suggested by @salihozdemir `useSelect` can have a function that returns a custom optionLabel and unique value from item to compare for duplicate as
```
useSelect({
    optionLabel: (item) => `${item.someField} ${item.anotherField}`,
    optionValue: (item) => item.foo.id,
})
```

Conqxeror:
## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

## What is the new behavior?

fixes #4880 

## Notes for reviewers

 ‼️ If provided changes meet the required changes than I should go further with docs. ‼️


BatuhanW:
Hey @Conqxeror thanks for the PR with great description as always!

Let's do the implementation according to the comment here: https://github.com/refinedev/refine/issues/4880#issuecomment-1742600902

`optionValue` and `optionLabel` params are currently string only, let's make each of them a union of string or function. And invoke the given function with `item` as in your implementation.

It would be amazing to implement similar changes for similar hooks other than `@refinedev/core`'s `useSelect`.

`@refinedev/antd` has following hooks which can be refactored in a similar way:
- useSelect
- useRadioGroup
- useCheckboxGroup

Conqxeror:
> Hey @Conqxeror thanks for the PR with great description as always!
> 
> Let's do the implementation according to the comment here: [#4880 (comment)](https://github.com/refinedev/refine/issues/4880#issuecomment-1742600902)
> 
> `optionValue` and `optionLabel` params are currently string only, let's make each of them a union of string or function. And invoke the given function with `item` as in your implementation.
> 
> It would be amazing to implement similar changes for similar hooks other than `@refinedev/core`'s `useSelect`.
> 
> `@refinedev/antd` has following hooks which can be refactored in a similar way:
> 
> * useSelect
> * useRadioGroup
> * useCheckboxGroup

Thanks, @BatuhanW ! Your feedback is always appreciated. I'll update `optionValue` and `optionLabel` as per the comment in [#4880 (comment)](https://github.com/refinedev/refine/issues/4880#issuecomment-1742600902). Also, extending these changes to `@refinedev/antd`'s `useSelect`, `useRadioGroup`, and `useCheckboxGroup` sounds great.

Conqxeror:
Hello @BatuhanW I wanted to know if all the required changes are satisfied or not. If so then I can go further with the docs. 

Review it whenever you are free, its okay if you take time.

packages/core/src/hooks/useSelect/index.ts
```diff
@@ -38,12 +38,16 @@ export type UseSelectProps<TQueryFnData, TError, TData> = {
    * Set the option's value
    * @default `"title"`
    */
-  optionLabel?: keyof TData extends string ? keyof TData : never;
+  optionLabel?: keyof TData extends string
+    ? keyof TData
+    : (item: TData) => string;
```
aliemir, line 43: The type doesn't seems correct. Here's a demonstration and the corrected type example [here in TS playground](https://www.typescriptlang.org/play?#code/PTAEBcE8AcFNQDIEMBGsA2AeAKgESeEqLAB7iwB2AJgM6gBKsAxgPYBOVmN4bAlhQHMANKCQVIAPlABeBs3adufQSLGSpsgNaxILAGag8BIqXLU6S-gIBQIUPYcB+UNt0GjhW2Af2AXPYAKXnIAW38PJABKGSlLQQBua2soOERUDBx8QmIySlo5Vg4uHitVcQ0ChWLlYVFyioDXfUMsk1zzUDibH1BnJvdW6x7-ClgAN1g2aIAfUACg0PDW6OlYksFIxOSYeAAFNhYqAFcmcBlQAG8hh14qEaOQtDZEn3Bg9Fh-LpeHJgJYATsSD+C6gW73R6TUAAXx+9iosBoTD40DeLAoX3WAjhnUI4CONH8ACJoEcUOheDQABawKhE0CzIlsWAAK2Y5DpDNARKobCQenARMS0KSrAo3FA6HS6H8yDQWH2hxO4AqRLe4A+QqSdjFEql8oATLLpZhFcdTqq9CwWELQHZJgc2NqwLqzvqMABmY3y00Hc0q84BaArKTQAB0CKRKLRFHidrALE0QA)

packages/antd/src/hooks/fields/useCheckboxGroup/index.ts
```diff
@@ -33,6 +33,18 @@ type UseCheckboxGroupProps<TQueryFnData, TError, TData> = Omit<
...
+   */
+  optionValue?: keyof TData extends string
+    ? keyof TData
+    : (item: TData) => string;
+  /**
+   * Set the option's label value
+   */
+  optionLabel?: keyof TData extends string
+    ? keyof TData
+    : (item: TData) => string;
```
aliemir, line 47: No need to add these here they will be provided by the `UseSelectProps` of `@refinedev/antd` and use select props from `@refinedev/core`. Same also applies to `useRadioGroup`

packages/core/src/hooks/useSelect/index.ts
```diff
@@ -301,9 +327,10 @@ export const useSelect = <
       setSearch([]);
       return;
     }
+
     setSearch([
       {
-        field: optionLabel,
+        field: optionLabel as string,
```
aliemir, line 333: I think we can add a new prop named `searchLabel` and make `optionLabel` the fallback value if its a string. Default value will be `"title"` (like `optionLabel`) but if a `searchLabel` is passed then it will use it (should be `key TData`).

something like this:

```suggestion
        field: searchLabel ?? (typeof optionLabel === "string" ? optionLabel : "title")
```

and ensure that we're prioritizing `searchLabel` and then `optionLabel` with type string  and if not, fallback to default `optionLabel` which is `"title"`. 

But before making any changes, let's validate this thought 😅 @alicanerdurmaz @BatuhanW

While we're adding `searchLabel`, I think it's best to also add `searchOperator` which is now hardcoded to `"contains"`. When using a custom `optionLabel` with function type, these don't really make sense to me to use.

aliemir (changes requested):
Thank you for the PR @Conqxeror! I've left couple of comments for the changes required.

packages/core/src/hooks/useSelect/index.ts
```diff
@@ -35,15 +35,24 @@ export type UseSelectProps<TQueryFnData, TError, TData> = {
...
    */
-  optionValue?: keyof TData extends string ? keyof TData : never;
+  optionValue?: keyof TData extends string
+    ? keyof TData
+    : never | ((item: TData) => string);
+  /**
+   * Set search label
+   * @default `"title"``
+   */
+  searchLabel?: keyof TData extends string ? keyof TData : never;
```
alicanerdurmaz, line 55: I don't think naming and JSDoc is descriptive here
Maybe It can be something like this

```suggestion
  /**
   * Field name to search for
   * @default `"title"``
   */
  fieldToSearch?: keyof TData extends string ? keyof TData : never;
```

alicanerdurmaz (approved):

packages/core/src/hooks/useSelect/index.ts
alicanerdurmaz, line 55: I don't think naming and JSDoc is descriptive here
Maybe It can be something like this

```suggestion
  /**
   * Field name to search for
   * @default `"title"``
   */
  fieldToSearch?: keyof TData extends string ? keyof TData : never;
```

alicanerdurmaz (changes requested):

packages/core/src/hooks/useSelect/index.ts
BatuhanW: Nice catch. Maybe `searchField` ?

alicanerdurmaz (approved):

aliemir (approved):

PR #5516, opened 2024-01-15
fix(strapi-v4): Strapi-v4-data-provider-not-working-with-nested-conditional-operators

vidurkhanal:
fix: Looks into solving [Issue #5479](https://github.com/refinedev/refine/issues/5479)

Explanation:

Current implementation of an utility function in strapi-v4 package, generateFilter(), doesn't handle all the edge cases, as pointed out by the issue owner.  The original code doesn't handle the case where the value of a conditional filter may be another conditional filter. It assumes that value of a conditional filter will be a logical filter.

I have chosen a recursive parsing approach to solve this and do believe produce expected raw string. However, the raw string after being parsed and stringified through "qs" adds unnecessary square braces to the URL params. Attached is the screenshot of the issue. The console.log is me logging the raw string before parsing and stringifying. 

Screenshot 2024-01-15 at 10 00 43 AM

### Test plan (required)

WIP: Added a new test case when there are recursive conditional filters.

### Closing issues
closes #5479 

### Self Check before Merge

Please check all items below before review.

-   [.] Corresponding issues are created/updated or not needed
-   [.] Docs are updated/provided or not needed
-   [.] Examples are updated/provided or not needed
-   [.] TypeScript definitions are updated/provided or not needed
-   [.] Tests are updated/provided or not needed
-   [.] Changesets are provided or not needed

PR #5694, opened 2024-02-27
fix(strapi-v4): error when using nested conditional operators

### LINKED ISSUE #5479: [BUG] Strapi v4 data provider does not working with nested filters $or and $and operator
### Describe the bug

image
image

### Steps To Reproduce

```ts
import { generateFilter } from "@refinedev/strapi-v4";

const filters = [
  {
    operator: "or",
    value: [
      {
        field: "affect_global",
        operator: "eq",
        value: true,
      },
      {
        operator: "and",
        value: [
          {
            field: "studio.id",
            operator: "eq",
            value: 2,
          },
          {
            field: "is_primary",
            operator: "eq",
            value: true,
          },
        ],
      },
    ],
  },
];

const query = generateFilter(filters);
```

### Expected behavior

`filters[$or][0][affect_global][$eq]=true&filters[$or][1][$and][0][studio][id][$eq]=2&filters[$or][1][$and][1][is_primary][$eq]=true`

### Screenshot

_No response_

### Desktop

_No response_

### Mobile

_No response_

### Additional Context

_No response_

rilrom:
## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

The original code doesn't handle the case where the value of a conditional filter may be another conditional filter. It assumes that the value of a conditional filter will be a logical filter.

## What is the new behavior?

Nested conditional filters are now handled appropriately.

fixes #5479 

## Notes for reviewers

The work for this fix was originally conducted in [#5516](https://github.com/refinedev/refine/pull/5516) however it was closed due to inactivity.

Something to keep in mind is that the `depth` value added to `parsedQuery` will only work for our current test cases. If a user were to create a more deeply nested filter object the filter would not work correctly.

We need to determine whether we want to allow `depth` to be updated by the user and/or whether we should mention this caveat in the documentation.


BatuhanW:
Hey @rilrom thanks for looking into this issue!

> We need to determine whether we want to allow depth to be updated by the user and/or whether we should mention this caveat in the documentation.

What would be the consequences of having more generous depth? Like 10, 20?

I think we should have a sensible default value, but we could also accept an optional parameter to set this depth even more, if someone really needs it.

Nevertheless, I believe we should mention this one in documentation.

BatuhanW (changes requested):
Thanks for the PR! 🚀  Let's get those lint errors fixed before we proceed.

rilrom:
> What would be the consequences of having more generous depth? Like 10, 20?

According to the `qs` documentation, the reason for the depth parameter is to help mitigate abuse when used to parse user input, so it's recommended to keep it at a reasonably small number.

I think 15 should be more than enough, I'd be very surprised if a user required a more deeply nested filter than that. Thinking more about it I'd be comfortable to omit this as a parameter and the documentation for now to minimise complexity (since it would only be there for extreme circumstances).

What are your thoughts on this @BatuhanW?

BatuhanW:
> > What would be the consequences of having more generous depth? Like 10, 20?
> 
> According to the `qs` documentation, the reason for the depth parameter is to help mitigate abuse when used to parse user input, so it's recommended to keep it at a reasonably small number.
> 
> I think 15 should be more than enough, I'd be very surprised if a user required a more deeply nested filter than that. Thinking more about it I'd be comfortable to omit this as a parameter and the documentation for now to minimise complexity (since it would only be there for extreme circumstances).
> 
> What are your thoughts on this @BatuhanW?

Hey @rilrom, I agree. Let's have 15 for now, which should be more than enough. We can consider accepting parameter only if we get requests.

BatuhanW (approved):
Great stuff 🚀 LGTM 👍🏼 cc @alicanerdurmaz @aliemir

aliemir (approved):

PR #5714, opened 2024-03-04
refactor(core): useForm improvements

### LINKED ISSUE #5460: [BUG] useForm onFinishUpdate promise never settles when in pessimistic mutationMode

### LINKED ISSUE #5702: [REFACTOR] useForm hook callbacks and mutations
We'll make some refactoring in `@refinedev/core`'s `useForm` hook to make the current codebase more readable and clean. 

While doing so there will be some changes made to the logic to make it simple and easy to follow both for the users and the maintainers.

Discussed changes is listed below but not might be more to be added;

- Resolve and reject conditions for the `onFinish` and being able to await for the mutation in pessimistic mode,
- `onMutationSuccess` and `onMutationError` will be triggered by the response of the mutation regardless of the mutation mode,
- More thorough tests both unit and e2e,
- Better integration with the devtools monitor,
- Simpler auto save logic with no separate submission block,
- Appropriate values and errors should be provided to the callbacks and promises,
- Avoiding any hanging promises (e.g. when mutation is errored out and `onMutationError` is provided).
- There might be changes required for the extended versions of the `useForm` hook in integration packages and they will also be addressed in the related PRs.

aliemir:
Refactored the internal logic of `useForm` to be more efficient and readable, along with few bug fixes and improvements to the `useForm` hook.

These changes are related to the issue [#5702](https://github.com/refinedev/refine/issues/5702)

- ~~Due to the confusion it causes by the name, `onFinish` is deprecated and replaced with `submit`.~~
- ~~Same also applies to `onFinishAutoSave` which is replaced with `submitAutoSave`.~~
- `onFinish` now rejects when; - `values` is not provided, - `resource` is not defined, - `id` is required but not provided.
  previously these cases were silently ignored.
- Same changes also applies to `onFinishAutoSave`.
- `onFinishAutoSave` had an issue with returning the appropriate promise after being called. This resulted in unhandled promise rejections and uncontrollable resolved promises. Now it is fixed, `onFinishAutoSave` will resolve and reject based on the response of the mutation.
- When using auto save, debounced calls will now be cancelled and the respective promises will be rejected with `"cancelled by debounce"` message. These changes might require an update to the code bases that uses `onFinishAutoSave` to handle the rejection of the promise to avoid unhandled promise rejections. 
- Combined the separated submit functions into one for sake of simplicity and consistency. (internal)
- `onFinish` rejects and resolved regardless of the `onMutationSuccess` and `onMutationError` hooks are provided or not. (Resolves [#5460](https://github.com/refinedev/refine/issues/5460))
- `meta` values were concatenated multiple times causing confusion and unexpected behavior, now it is fixed. (internal)
- Moved the `id` determination/inference logic to a separate hook. (internal)

## PR Checklist

- [x] Related issue(s) linked

## Issues

Resolves #5702 
Fixes #5460


packages/core/src/hooks/form/index.ts
```diff
@@ -1 +1,341 @@
...
+  const isCreate = action === "create";
+
+  const combinedMeta = getMeta({
+    resource,
+    meta: pickedMeta,
+  });
+
+  const idRequired = (isEdit || isClone) && Boolean(props.resource);
+  const idHandled = typeof props.id !== "undefined";
+  const idDismissed = props.queryOptions?.enabled === false;
```
BatuhanW, line 112:
```suggestion
  const isQueryDisabled = props.queryOptions?.enabled === false;
```

packages/core/src/hooks/form/index.ts
```diff
@@ -1 +1,341 @@
...
+  const isClone = action === "clone";
+  const isCreate = action === "create";
+
+  const combinedMeta = getMeta({
+    resource,
+    meta: pickedMeta,
+  });
+
+  const idRequired = (isEdit || isClone) && Boolean(props.resource);
+  const idHandled = typeof props.id !== "undefined";
```
BatuhanW, line 111:
```suggestion
  const hasIdProp = typeof props.id !== "undefined";
```

packages/core/src/hooks/form/index.ts
```diff
@@ -1 +1,341 @@
...
+  const idDismissed = props.queryOptions?.enabled === false;
+
+  /**
+   * When a custom resource is provided through props, `id` will not be inferred from the URL to avoid any potential faulty requests.
+   * In this case, `id` is required to be passed through props.
+   * If `id` is not handled, a warning will be thrown in development mode.
+   */
+  warnOnce(
+    idRequired && !idHandled && !idDismissed,
```
BatuhanW, line 120:
```suggestion
    idRequired && !hasIdProp && !isQueryDisabled,
```

With the suggested changes above, I think this condition becomes easier to read.

packages/core/src/hooks/form/index.ts
```diff
@@ -1 +1,341 @@
...
+  });
+
+  // biome-ignore lint/correctness/useExhaustiveDependencies: This is a controlled unmounting effect.
+  React.useEffect(() => {
+    // After `autosaved` is set to `true`, it won't be set to `false` again.
+    // Therefore, the `invalidate` function will be called only once at the end of the hooks lifecycle.
+    return () => {
+      if (props.autoSave?.invalidateOnUnmount && autosaved) {
+        invalidate({
+          id,
```
BatuhanW, line 189: Should we check existence of `id` here?

packages/core/src/hooks/form/index.ts
```diff
@@ -1 +1,341 @@
...
+    const onSuccessRedirect = (id?: BaseKey) => redirect(redirectAction, id);
+
+    const submissionPromise = new Promise<
+      // biome-ignore lint/suspicious/noConfusingVoidType: Void is an expected case for this promise.
+      CreateResponse<TResponse> | UpdateResponse<TResponse> | void
+    >((resolve, reject) => {
+      // Reject the mutation if the resource is not defined
+      if (!resource) return reject(missingResourceError);
+      // Reject the mutation if the `id` is not defined in edit action
+      if (isEdit && !id) return reject(missingIdError);
```
BatuhanW, line 214: I think we shouldn't reject here. AFAIk, it wasn't rejected before. And users may have a case where they update resource without an id e.g.: `api/users/settings`.

BatuhanW:
```suggestion
      if (isClone && !id) return reject(missingIdError);
```

I think we should fail if clone is missing id tho.

packages/core/src/hooks/form/index.ts
```diff
@@ -1 +1,341 @@
...
+      if (!values) return reject(missingValuesError);
+
+      if (!isPessimistic && !isAutosave) {
+        // If the mutation mode is not pessimistic, handle the redirect immediately in an async manner
+        // `setWarnWhen` blocks the redirects until set to `false`
+        // If redirect is done before the value is properly set, it will be blocked.
+        // We're deferring the execution of the redirect to ensure that the value is set properly.
+        deferExecution(() => onSuccessRedirect());
+        // Resolve the promise immediately
+        resolve();
```
BatuhanW, line 225: We can resolve with mutation variables for optimistic update.

packages/core/src/hooks/form/index.ts
BatuhanW: I think we can also use `id` variable here, coming from `useId` hook.

packages/core/src/hooks/form/index.ts
```diff
@@ -1 +1,341 @@
...
+
+  const {
+    resource,
+    action: routeAction,
+    identifier,
+  } = useResource(props.resource);
+
+  const [autosaved, setAutosaved] = React.useState(false);
+
+  const [id, setId] = useId({ id: props.id, resource: props.resource });
```
BatuhanW, line 90:
```suggestion
  const [id, setId] = useId({ id: props.id, resource: props.resource });
```

Do we handle the case, `resource` string is not the name of the string, but it's `identifier` of the resource?

omeraplak (approved):

PR #5744, opened 2024-03-13
feat(mui): editable Data Grid

### LINKED ISSUE #5656: [FEAT] Editable MUI Data Grid

There is currently no quick way to implement editable MUI Data Grid. 

There is alternative of using useEditableTable in Ant Design System. 

### Additional context

_No response_

It would be very helpful if editable Data Grid could be set up quickly, like Editable Table in Ant.

beg1c:
## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

## What is the new behavior?

Data Grid can now be edited by specifying `editable` property on DataGrid column definition, just as in MUI Data Grid documentation. `useDataGrid` will now utilize existing Data Grid props `processRowUpdate`, `onRowEditStart` and `onCellEditStart` to update resource whenever `processRowUpdate` is triggered. 

fixes #5656 (issue)

## Notes for reviewers

This is lightweight implementation of this functionality, using only existing Data Grid props. This functionality can be later extended for handling multiline editing.

`processRowUpdate` must return value to Data Grid, that is why we must return oldRow if mutation does not succeed.

I've set title column on `examples/table-material-ui-use-data-grid` as editable.


beg1c:
Some problems occurred because my fork was not completely synced. Resolved it now.

examples/table-material-ui-use-data-grid/src/pages/posts/list.tsx
```diff
@@ -45,7 +45,13 @@ export const PostList: React.FC = () => {
...
         width: 50,
       },
-      { field: "title", headerName: "Title", minWidth: 400, flex: 1 },
+      {
+        field: "title",
+        headerName: "Title",
+        minWidth: 400,
+        flex: 1,
+        editable: true,
+      },
```
BatuhanW, line 54: We have cypress e2e tests for this example, let's update them to test this new behavior.

packages/mui/src/hooks/useDataGrid/index.ts
```diff
@@ -263,6 +267,13 @@ export function useDataGrid<
     };
   };
 
+  const edit = useForm<TQueryFnData, TError, TData>({
+    action: "edit",
+    redirect: false,
+  });
```
BatuhanW, line 273: Users may provide custom resource, let's make sure we handle that and other similar props if any.

packages/mui/src/hooks/useDataGrid/index.ts
```diff
@@ -263,6 +267,13 @@ export function useDataGrid<
...
+
+  const { setId, onFinish: onFinishEdit } = edit;
```
BatuhanW, line 275: I think this is useless, better to destruct from hook.

packages/mui/src/hooks/useDataGrid/index.ts
```diff
@@ -310,6 +321,16 @@ export function useDataGrid<
...
+      processRowUpdate: async (newRow, oldRow) => {
+        try {
+          await onFinishEdit(newRow);
+          return newRow;
+        } catch {
+          return oldRow;
+        }
+      },
+      onCellEditStart: (params) => setId(params?.id),
+      onRowEditStart: (params) => setId(params?.id),
```
BatuhanW, line 333: Let's add tests for the updates in this file.

BatuhanW (changes requested):
Hey @beg1c thanks for the PR! Added some comments.

packages/mui/src/hooks/useDataGrid/index.ts
beg1c: We will have to extend `UseDataGridReturnType` with `onFinish`, `setId`, `id` and maybe `formLoading` from `UseFormReturnType`. That will also allow users to completely override this functionality and call mutations on different events. 
I was not able to write tests without these return types as only way to call `setId` is to call `onCellEditStart` which accepts a lot of different arguments that I can not replicate.

beg1c:
@BatuhanW Added tests and some changes.  I'd appreciate it if you could take a look when you have a moment.  I plan to update the documentation later since there might be more changes before merging. Additionally, it seems we're lacking functionality where the user can call `useForm` with different `action` based on the call. This would enable users to extend the functionality and create new rows in the DataGrid, not just edit existing ones. Same goes for delete. I will look into it more when I catch time.

BatuhanW:
> @BatuhanW Added tests and some changes. I'd appreciate it if you could take a look when you have a moment. I plan to update the documentation later since there might be more changes before merging. Additionally, it seems we're lacking functionality where the user can call `useForm` with different `action` based on the call. This would enable users to extend the functionality and create new rows in the DataGrid, not just edit existing ones. Same goes for delete. I will look into it more when I catch time.

@beg1c appreciate the effort! Looks good initially, I'll review more in detail later. Also cc @aliemir @alicanerdurmaz

packages/mui/src/hooks/useDataGrid/index.ts
```diff
@@ -45,49 +48,62 @@ type DataGridPropsType = Required<
...
+      "defaultBehavior"
+    > & {
+      /**
+       * Default behavior of the `setFilters` function
+       * @default "replace"
+       */
+      defaultBehavior?: "replace" | "merge";
+    }
+  >;
+  formProps?: UseFormProps<TQueryFnData, TError, TData>;
```
alicanerdurmaz, line 92: maybe we should omit these props `"autoSave" | "action" | "redirect"` because `action` and `redirect` can't be overridden and `autoSave` doesn't  work because `processRowUpdate` is triggered `onBlur`


```suggestion
  formProps?: Omit<
    UseFormProps<TQueryFnData, TError, TData>,
    "autoSave" | "action" | "redirect"
  >;
```

@BatuhanW @aliemir @beg1c  What do you think?

alicanerdurmaz (approved):

packages/mui/src/hooks/useDataGrid/index.ts
alicanerdurmaz, line 92: maybe we should omit these props `"autoSave" | "action" | "redirect"` because `action` and `redirect` can't be overridden and `autoSave` doesn't  work because `processRowUpdate` is triggered `onBlur`


```suggestion
  formProps?: Omit<
    UseFormProps<TQueryFnData, TError, TData>,
    "autoSave" | "action" | "redirect"
  >;
```

@BatuhanW @aliemir @beg1c  What do you think?

alicanerdurmaz (changes requested):

packages/mui/src/hooks/useDataGrid/index.ts
beg1c: @alicanerdurmaz I agree!

beg1c:
@BatuhanW @alicanerdurmaz updated docs. Do you think we should accept boolean `editable` as prop to `useDataGrid` or is this PR over? For further usage I don't think we need that prop, but we could break older versions if we don't accept it. If some users have their own implementation of editable datagrid, and they didn't declare `processRowUpdate` on `<DataGrid />` component, `<DataGrid />` `processRowUpdate` will trigger our `useForm` `onFinish` every time column is commited.

alicanerdurmaz:
> @BatuhanW @alicanerdurmaz updated docs. Do you think we should accept boolean `editable` as prop to `useDataGrid` or is this PR over?
> 
> For further usage I don't think we need that prop, but we could break older versions if we don't accept it. If some users have their own implementation of editable datagrid, and they didn't declare `processRowUpdate` on `<DataGrid />` component, `<DataGrid />` `processRowUpdate` will trigger our `useForm` `onFinish` every time column is commited.
> 
> So for example if someone defined their data grid as:
> 
> ```
> <DataGrid
>     {...dataGridProps}
>     columns={columns}
>     onRowEditStart={ownSetIdFunction}
>     onRowEditCommit={ownMutateFunction}
> />
> ```
> 
> which is wrong as `processRowUpdate` should be used for migrations from `DataGrid version 6`, our `processRowUpdate` from `...dataGridProps` spread could break that functionality.
> 
> I don't know, what do you think?

I believe you are right. We should accept `editable` prop.

something like this:
```tsx
useForm({
  queryOptions: {
    enabled: editable,
  },
});

processRowUpdate: async (newRow, oldRow) => {
   if(!editable) return
   /* .... */
}
  
 ```

BatuhanW:
Hey @beg1c we've merged `releases/april` branch, could you re-open your PR?

PR #5762, opened 2024-03-17
fix(react-router-v6): wrong type declaration for legacy router at `/legacy` imports

aliemir:
Fixed the wrong type declaration for legacy router subexport at `/legacy`. Previously imports from `@refinedev/react-router-v6/legacy` used type declarations from new router, which caused type errors. Now it's fixed and the correct type declarations are used.


aliemir:
This PR is on hold in favor of #5765

aliemir:
Closing in favor of #5765

PR #5808, opened 2024-04-01
feat(core): button hooks

aliemir:
Added button hooks to unify the logic in the `@refinedev/core` package and use it in UI integrations via hooks. This will reduce code duplication between UI packages and provide a maintainable shared codebase for the buttons. Now users can also use these hooks to create their buttons for their own UI integrations.

New Hooks:

- `useListButton`: A navigation button that navigates to the list page of a resource.
- `useCreateButton`: A navigation button that navigates to the create page of a resource.
- `useShowButton`: A navigation button that navigates to the show page of a record.
- `useEditButton`: A navigation button that navigates to the edit page of a record.
- `useCloneButton`: A navigation button that navigates to the clone page of a record.
- `useRefreshButton`: A button that triggers an invalidation of the cache of a record.
- `useDeleteButton`: A button that triggers a delete mutation on a record.
- `useSaveButton`: A button to be used inside a form to trigger a save mutation.
- `useExportButton`: A button to be used with `useExport` to trigger an export bulk data of a resource.
- `useImportButton`: A button to be used with `useImport` to trigger an import bulk data for a resource.

We've also updated all the UI packages to use these hooks instead of the duplicated code blocks.

By moving to hooks from `@refinedev/core` we've also introduced a small change in the `id` behavior of the related buttons. Now the `id` value will inferred just like the `useForm` hooks, meaning if the explicitly set `resource` value is the same with the inferred `resource` value, inferred `id` from the router will also be used but if explicitly set `resource` is different from the inferred `resource`, we'll require `id` to be also set explicitly to work properly. 

## Bugs / Features

- [x] Related issue(s) linked

RK-582


omeraplak (approved):
⚡️

BatuhanW (approved):

PR #5851, opened 2024-04-15
refactor(core): prevent early access to `signal` from query context

### LINKED ISSUE #5843: [BUG] useSelect hook ignores staleTime
### Describe the bug

I'm facing the same problem described in https://github.com/refinedev/refine/issues/1815, where `useSelect `duplicates requests for Options list, but in this case it ignores `staleTime `option. 

### Steps To Reproduce

1. Go to https://codesandbox.io/p/devbox/charming-bohr-jt6tcl?file=%2Fsrc%2Fcomponents%2FResourceSelect.tsx%3A10%2C37
2. Edit post
3. Check the browser console's requests tab.

### Expected behavior

The options list should be loaded only once.

### Packages

"@refinedev/cli": "^2.16.21",
"@refinedev/core": "^4.47.1",
"@refinedev/devtools": "^1.1.32",
"@refinedev/kbar": "^1.3.6",
"@refinedev/simple-rest": "^5.0.1",
"@refinedev/antd": "^5.37.4",
"antd": "^5.0.5",

### Additional Context

I've created another simple component, `ResourceUseQuerySelect`, that uses just the `useQuery `hook and makes requests via `datasource.getList`. It seems to have the correct behavior for requesting data.
Both components are used on the same page and form.

Upd: I've tried to call `useList` hook in the component, and it duplicates requests too. It looks like there might be something wrong with the useList hook because, under the hood, it uses useQuery, which works fine.

I also tested in prod mode, and non strict mode.

aliemir:
In query hooks, `signal` was accessed directly by destructuring which was causing issues in development mode with duplicated requests. This change accesses `queryContext` instead of destructured `signal` properly to prevent `@tanstack/react-query` from setting `abortSignalConsumed` flag unexpectedly.

Resolves #5843

### DIFF
--- a/.changeset/eleven-kangaroos-hope.md
+++ b/.changeset/eleven-kangaroos-hope.md
@@ -1,0 +1,9 @@
+---
+"@refinedev/core": patch
+---
+
+refactor: prevented early accessing `signal` from `useQuery` of `@tanstack/react-query`
+
+In query hooks, `signal` was accessed directly by destructuring which was causing issues in development mode with duplicated requests. This change accesses `queryContext` instead of destructured `signal` properly to prevent `@tanstack/react-query` from setting `abortSignalConsumed` flag unexpectedly.
+
+Resolves [#5843](https://github.com/refinedev/refine/issues/5843)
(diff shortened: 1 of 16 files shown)


aliemir:
Tests will be updated accordingly. They're failing due to the previous patch done in e2e tests to workaround the issue resolved in this PR 😅

PR #5989, opened 2024-05-26
feat(mui): editable Data Grid #5744

lnikitadobrenkol:
## Bugs / Features
This PR introduces the ability to make MUI Data Grid columns editable by setting an editable property on specific column definitions.

Currently, the MUI Data Grid does not support inline editing directly through configuration.

## What is the new behavior?
With the changes in this PR, developers can now configure columns to be editable directly from the MUI Data Grid setup, enhancing the grid's flexibility for various use cases.

fixes # ([5656](https://github.com/refinedev/refine/issues/5656))

## Notes for reviewers
I have extensively tested the functionality introduced in this PR, primarily based on the initial implementation by @beg1c. After setting up a test project to integrate the enhanced dataGridHook, it has met the expected outcomes. I encourage a thorough review to ensure that all aspects of the new editable feature align with our project standards and user requirements. Please provide feedback or suggest improvements if you identify potential enhancements or adjustments. I am open to discussions and willing to make any necessary changes to perfect this feature.

### DIFF
--- a/packages/mui/src/hooks/useDataGrid/index.ts
+++ b/packages/mui/src/hooks/useDataGrid/index.ts
@@ -5,6 +5,9 @@ import {
   Pagination,
   pickNotDeprecated,
   Prettify,
+  useForm,
+  UseFormProps,
+  UseFormReturnType,
   useLiveMode,
   useTable as useTableCore,
   useTableProps as useTablePropsCore,
@@ -45,6 +48,9 @@ type DataGridPropsType = Required<
     | "disableRowSelectionOnClick"
     | "onStateChange"
     | "paginationMode"
+    | "processRowUpdate"
+    | "onCellEditStart"
+    | "onRowEditStart"
   >
 > &
   Pick<
@@ -52,42 +58,55 @@ type DataGridPropsType = Required<
     "paginationModel" | "onPaginationModelChange" | "filterModel"
   >;
 
-export type UseDataGridProps<TQueryFnData, TError, TSearchVariables, TData> =
-  Omit<
-    useTablePropsCore<TQueryFnData, TError, TData>,
-    "pagination" | "filters"
-  > & {
-    onSearch?: (data: TSearchVariables) => CrudFilters | Promise<CrudFilters>;
-    pagination?: Prettify<
-      Omit<Pagination, "pageSize"> & {
-        /**
-         * Initial number of items per page
-         * @default 25
-         */
-        pageSize?: number;
-      }
-    >;
-    filters?: Prettify<
-      Omit<
-        NonNullable<useTablePropsCore<TQueryFnData, TError, TData>["filters"]>,
-        "defaultBehavior"
-      > & {
-        /**
-         * Default behavior of the `setFilters` function
-         * @default "replace"
-         */
-        defaultBehavior?: "replace" | "merge";
-      }
-    >;
-  };
+export type UseDataGridProps<
+  TQueryFnData extends BaseRecord,
+  TError extends HttpError,
+  TSearchVariables,
+  TData,
+> = Omit<
+  useTablePropsCore<TQueryFnData, TError, TData>,
+  "pagination" | "filters"
+> & {
+  onSearch?: (data: TSearchVariables) => CrudFilters | Promise<CrudFilters>;
+  pagination?: Prettify<
+    Omit<Pagination, "pageSize"> & {
+      /**
+       * Initial number of items per page
+       * @default 25
+       */
+      pageSize?: number;
+    }
+  >;
+  filters?: Prettify<
+    Omit<
+      NonNullable<useTablePropsCore<TQueryFnData, TError, TData>["filters"]>,
+      "defaultBehavior"
+    > & {
+      /**
+       * Default behavior of the `setFilters` function
+       * @default "replace"
+       */
+      defaultBehavior?: "replace" | "merge";
+    }
+  >;
+  formProps?: Omit<
+    UseFormProps<TQueryFnData, TError, TData>,
+    "autoSave" | "action" | "redirect"
+  >;
+};
 
 export type UseDataGridReturnType<
-  TData extends BaseRecord = BaseRecord,
+  TQueryFnData extends BaseRecord = BaseRecord,
   TError extends HttpError = HttpError,
   TSearchVariables = unknown,
+  TData extends BaseRecord = TQueryFnData,
 > = useTableReturnTypeCore<TData, TError> & {
   dataGridProps: DataGridPropsType;
   search: (value: TSearchVariables) => Promise<void>;
+  formProps: Pick<
+    UseFormReturnType<TQueryFnData, TError, TData>,
+    "onFinish" | "setId" | "id" | "formLoading"
+  >;
 };
 
 /**
@@ -134,12 +153,13 @@ export function useDataGrid<
   metaData,
   dataProviderName,
   overtimeOptions,
+  formProps,
 }: UseDataGridProps<
   TQueryFnData,
   TError,
   TSearchVariables,
   TData
-> = {}): UseDataGridReturnType<TData, TError, TSearchVariables> {
+> = {}): UseDataGridReturnType<TQueryFnData, TError, TSearchVariables, TData> {
   const theme = useTheme();
   const liveMode = useLiveMode(liveModeFromProp);
 
@@ -263,6 +283,16 @@ export function useDataGrid<
     };
   };
 
+  const { setId, onFinish, id, formLoading } = useForm<
+    TQueryFnData,
+    TError,
+    TData
+  >({
+    ...formProps,
+    action: "edit",
+    redirect: false,
+  });
+
   return {
     tableQueryResult,
     dataGridProps: {
@@ -310,6 +340,16 @@ export function useDataGrid<
           )}`,
         },
       },
+      processRowUpdate: async (newRow, oldRow) => {
+        try {
+          await onFinish(newRow);
+          return newRow;
+        } catch {
+          return oldRow;
+        }
+      },
+      onCellEditStart: (params) => setId(params?.id),
+      onRowEditStart: (params) => setId(params?.id),
     },
     current,
     setCurrent,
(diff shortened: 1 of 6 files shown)


aliemir (changes requested):
Hey @lnikitadobrenkol thank you for the PR! Do you think using `useForm` has any advantages over `useUpdate` in this specific case? Since `useForm` features are mostly disabled and action is hardcoded to `"edit"`, I think we can use `useUpdate` and  have a simpler implementation by omitting `id` and `setId` 🤔

lnikitadobrenkol:
Hello @aliemir 
I've been working on enhancing the useDataGrid hook to handle errors and state more robustly, specifically around reverting to original data when updates fail. However, I’m stuck with an issue where, despite the update operation being rejected (we simulate this in tests), the function still returns the modified data instead of the original.

Here’s the quick rundown:
Function: processRowUpdate
Expected: Return old data on update failure.
Actual: It returns the new, incorrect data.

I’ve spent quite a few hours on this and tried various fixes, but no luck so far. Could you please help me with this – any commits, examples, ideas or suggestions would be greatly appreciated!
We can arrange a short call if needed - so I can share the screen to show what exactly I am doing, (maybe I completely on the wrong direction)

chatpong-v (changes requested):
To update feature

lnikitadobrenkol:
@BatuhanW @aliemir Hey guys! I need your help in implementing useUpdate. Please have a look

aliemir:
Hey @lnikitadobrenkol i'll try to check your comment today and get back to you 🙏

lnikitadobrenkol:
Hi @aliemir,I think I was finally able to resolve the issue and updated the PR :space_invader:! Please review the latest commit and let me know if further adjustments are needed. I would  appreciate your feedback to ensure everything meets the standards

packages/mui/src/hooks/useDataGrid/index.ts
```diff
@@ -45,41 +47,50 @@ type DataGridPropsType = Required<
     | "disableRowSelectionOnClick"
     | "onStateChange"
     | "paginationMode"
+    | "processRowUpdate"
   >
 > &
   Pick<
     DataGridProps,
     "paginationModel" | "onPaginationModelChange" | "filterModel"
```
aliemir, line 55:
```suggestion
  >
> &
  Pick<
    DataGridProps,
    "paginationModel" | "onPaginationModelChange" | "filterModel" | "processRowUpdate"
```

Let's move this here, I don't think this needs to be present for all `useDataGrid` usages. Placing it here will mark it as optional.

I think we can add `editable: boolean` prop to `useDataGrid` and make a conditional return for `processRowUpdate`. I'm not sure but sending this even without any editable columns may cause issues with current users implemented editable mode in their codebase.

packages/mui/src/hooks/useDataGrid/index.ts
```diff
@@ -45,41 +47,50 @@ type DataGridPropsType = Required<
...
-         * Default behavior of the `setFilters` function
-         * @default "replace"
-         */
-        defaultBehavior?: "replace" | "merge";
-      }
-    >;
-  };
+export type UseDataGridProps<
+  TQueryFnData extends BaseRecord,
+  TError extends HttpError,
```
aliemir, line 60: Is this change required? Not sure where this is useful 🤔

packages/mui/src/hooks/useDataGrid/index.ts
```diff
@@ -45,41 +47,50 @@ type DataGridPropsType = Required<
...
+       * Default behavior of the `setFilters` function
+       * @default "replace"
+       */
+      defaultBehavior?: "replace" | "merge";
+    }
+  >;
+  formProps?: Omit<
+    UseFormProps<TQueryFnData, TError, TData>,
+    "autoSave" | "action" | "redirect"
+  >;
```
aliemir, line 92: We don't need that since we don't use `useForm` anymore in this implementation. But we can add `updateMutationOptions` (from `UseFormProps`) similar to what we do in `useForm` implementation of `@refinedev/core`. We can pass this `updateMutationOptions` to `useUpdate`. Let me know if help is needed to do this.

packages/mui/src/hooks/useDataGrid/index.ts
```diff
@@ -263,6 +279,33 @@ export function useDataGrid<
...
+              resolve(data);
+            },
+          },
+        );
+      });
+      return newRow;
+    } catch (error) {
+      return oldRow;
+    }
+  };
```
aliemir, line 307: Can we use `React.useCallback` here? Let's make sure we don't cause any unwanted re-renders for this prop 😅

packages/mui/src/hooks/useDataGrid/index.ts
```diff
@@ -263,6 +279,33 @@ export function useDataGrid<
...
+          {
+            onError: (error) => {
+              reject(error);
+            },
+            onSuccess: (data) => {
+              resolve(data);
+            },
+          },
+        );
+      });
```
aliemir, line 302: Check out the docs from Material UI [here](https://mui.com/x/react-data-grid/editing/#server-side-validation). Returning old or new data will close the edit state and update the row (if new data is returned). Rejecting the promise will keep the edit state. 

Let's update this code with returning a promise, `onSuccess` can `resolve(newRow)` and `onError` can `reject(error)`. By default if user has provided any notification provider, Refine's `useUpdate` will display an error notification. In this case, it will be better to keep the edit state rather than closing it with old data.

packages/mui/src/hooks/useDataGrid/index.ts
```diff
@@ -310,6 +353,7 @@ export function useDataGrid<
           )}`,
         },
       },
+      processRowUpdate,
```
aliemir, line 356: As I mentioned in the above comments, we can return this conditionally by looking for the `editable: boolean` prop.

packages/mui/src/hooks/useDataGrid/index.ts
```diff
@@ -325,5 +369,9 @@ export function useDataGrid<
     search,
     createLinkForSyncWithLocation,
     overtime,
+    formProps: {
+      processRowUpdate,
+      formLoading,
+    },
```
aliemir, line 375: Mentioned above as well but we can remove this.

aliemir (changes requested):
Hey @lnikitadobrenkol thank you for the updates! Looks great! I've just left couple of comments to improve the usage and remove some unnecessary stuff. Can you work on those? 

I'll also try to check the e2e tests and provide a review for it 🚀 

Thanks again for the great work 👏 👏

lnikitadobrenkol:
@aliemir Thank you for the detailed review!! It really helps, appreciate it. Sure, I'll have a look at the comments and commit the fix  as soon as possible for me

packages/mui/src/hooks/useDataGrid/index.ts
lnikitadobrenkol: I removed formProps and added updateMutationOptions, but please have a look, and if it is not what you meant - the help is needed to do this

lnikitadobrenkol:
@aliemir I updated the PR based on your feedback. Please have a look and and let me know if we can proceed or more changes are required

lnikitadobrenkol:
@aliemir @BatuhanW @omeraplak Hey guys! I think I found the way to fix cypress e2e tests, just made the commit. Could you please review the PR - and let's fix any problem that are left, and proceed with this issue further?

packages/mui/src/hooks/useDataGrid/index.ts
```diff
@@ -79,6 +83,10 @@ export type UseDataGridProps<TQueryFnData, TError, TSearchVariables, TData> =
         defaultBehavior?: "replace" | "merge";
       }
     >;
+    editable?: boolean;
+    updateMutationOptions?: {
+      retry?: number;
+    };
```
aliemir, line 89: I think we can use `UseUpdateProps` from `@refinedev/core` but it might not be exported from the package 😅 Maybe we can export it from `src/hooks/data/index.ts` and use it in here for typing;

```ts
updateMutationOptions?: UseUpdateProps<
    TData,
    TError,
    TData
  >["mutationOptions"];
```

packages/mui/src/hooks/useDataGrid/index.ts
```diff
@@ -263,6 +272,32 @@ export function useDataGrid<
     };
   };
 
+  const { mutate, isLoading: formLoading } = useUpdate();
```
aliemir, line 275: We can also use generics here just like I've mentioned above for the `updateMutationOptions` 🤔 

```suggestion
  const { mutate, isLoading: formLoading } = useUpdate<TData, TError, TData>();
```

aliemir (changes requested):
Left two small comments that we can work on. Thank you for the changes, e2e test looks great! I'll also try to test this feature myself in my local 🤞 

Let me know if you can work on the comments, if not I can try to take a look at them myself 🙏

lnikitadobrenkol:
Thank you! I'll have a look at the comments and work on the fix this week.
Sure, please let me know how you find this feature locally and if there is anything we can improve

lnikitadobrenkol:
@aliemir hello! I implemented the changes you suggested, please have a look

aliemir (approved):
Hey @lnikitadobrenkol found some issues with the implementation and the example was using a custom `processRowUpdate` rather than the one from `useDataGrid`, was that intentional? 

Updated the code for `processRowUpdate` and updated the example code. While doing so, also had a small update in the core package to export the required types and updated the type import for `UseUpdateProps`.

lnikitadobrenkol:
Hey @aliemir!
No, it was not that intentional - I guess I missed it somehow.
Thank you for spotting it, and doing other updates - looks great!

lnikitadobrenkol:
@aliemir What do you think - are we ready to merge the PR or there is something else left to work on?

lnikitadobrenkol:
@aliemir @BatuhanW
Could you please review this PR whenever it's convenient for you and let me know if it's ready to be merged? I'd love to include this feature in the July release. I'll be on vacation from June 22-29, so if any changes are needed, it would be great to address them this week.

aliemir:
Hey @lnikitadobrenkol We'll merge this PR to be included in our next release. In the context of this feature, we don't require any additional changes, thank you for your contribution 🙏

Issue #6096, opened 2024-07-03
[BUG]: incorrect behavior on debonuce of `onSearch` in `useSelect`


Dominic-Preap:
### Describe the bug

When use `onSearch` in `useSelect`, we see that fetch requests called repeatedly after delay on `debonuce` value. Video demo is below.

https://github.com/refinedev/refine/assets/14802170/11cce211-1339-437b-96e8-a2bcbc2f9586

### Steps To Reproduce

Sample code:

```tsx
const { onSearch, queryResult } = useSelect({
  resource: 'test',
  debounce: 1500,
  searchField: 'search'
});

return <input onChange={e=>onSearch(e.currentTarget.value)}/>
```

### Expected behavior

 It should delay then executed `onSearch` once or throttling `onSearch` function. After checking in package it seems to use `debounce` from loadash. I suggest using [use-debounce](https://github.com/xnimorz/use-debounce) or any libs that are similar.  

### Packages

- @refinedev/core

### Additional Context

This will lead to multiple requests to the server if not handle properly.

aliemir:
Hey @Dominic-Preap, I've tested the behavior in our test environment and also used your code in a project but could not reproduce the issue. Looks like `onSearch` is working as expected in both cases. Can you please provide a small reproducible example?

Dominic-Preap:
Let me add testing repo.

Dominic-Preap:
Here is the test [repo](https://github.com/Dominic-Preap/refine-test/blob/master/src/pages/blog-posts/create.tsx#L130). I think I know the problem is. We use `Controller` from hook-from and render autocomplete input inside & not using `useCallback` that cause it fetch multiple. I'm not sure if it should handle by user or refine.

```tsx 
// blog-posts/create.tsx

// Page
<Controller
  control={control}
  name="patient.name"
  render={({ field: { onChange, value }, fieldState: { error } }) => (
    <CategorySearch
      error={error?.message}
      onChange={onChange}
      value={value}
    />
  )}


const CategorySearch: React.FC<Props> = ({ onChange, value, error }) => {
  const { onSearch, queryResult } = useSelect({
    resource: "categories",
    debounce: 1500,
    searchField: "title",
  });

  // ! Debounce working okay <---- SWITCH FUNCTION HERE
 const handleChange = useCallback((value: string) => {
    onChange?.(value);
    onSearch(value);
  }, []); 

  // ! Debounce run fetch multiple <---- SWITCH FUNCTION HERE
[...]

PR #6125, opened 2024-07-10
fix(core): update debounce behavior on `onSearch` in `useSelect`

Dominic-Preap:
## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

Debounce behavior on `onSearch`  is not working correctly when use inside `Controller` of react-hook-form causing re-fechting  multiple times.

## What is the new behavior?

Update `onSearch` to memoized value to prevent from re-fetching instead on return as inlined.

## Notes for reviewers

fix(core): update debounce behavior on `onSearch` in `useSelect`

Fixes https://github.com/refinedev/refine/issues/6096

### DIFF
--- a/.changeset/thin-adults-complain.md
+++ b/.changeset/thin-adults-complain.md
@@ -1,0 +1,9 @@
+---
+"@refinedev/core": minor
+---
+
+fix: update debounce behavior on `onSearch` in `useSelect`
+
+Now debounce behavior is working correctly on `onSearch` in `useSelect` when using inside `Controller` of react-hook-form.
+
+Resolves [#6096](https://github.com/refinedev/refine/issues/6096)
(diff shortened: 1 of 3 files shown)


BatuhanW (commented):
Hey @Dominic-Preap thanks for the PR! Checked it briefly and it looks good. Could you also add tests? Then we can do a detailed review. 🚀

Dominic-Preap:
@BatuhanW There are already test case on debounce function. My case is using `useSelect` inside `Controller` in react-hook-form, so I'm not familiar on how to write test case with that one. Could anyone can help on that?

BatuhanW:
Hey @Dominic-Preap thanks for the clarification, we'll check it!

aliemir (approved):
Hey @Dominic-Preap, thank you for the PR! 🚀

I've updated one test case which was not responding to the changes in the `onSearch` prop because it lacked `rerender()` calls. Also added one more case to ensure `onSearch` prop can change without breaking the `debounce` interval. I think these are enough to test out the initial issue we're fixing here. 

Thanks again for the contribution! This will be included in our next release 👏 👏

PR #6133, opened 2024-07-12
feat(nestjs-query): add support to custom get requests

igordonin:
## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

Custom requests do not make use of the params that are passed to it, including the HTTP Request Method, the URL and the headers.

## What is the new behavior?

Now it uses all of the custom params and correctly supports POST and GET Request Methods, throws errors for unsupported methods, and also uses the custom URL and headers passed to the function.

## Notes for reviewers

Not at this time


packages/nestjs-query/src/dataProvider/index.ts
```diff
@@ -419,30 +419,38 @@ const dataProvider = (client: GraphQLClient): Required<DataProvider> => {
...
 
       if (meta?.rawQuery) {
-        const response = await client.request<BaseRecord>(
-          meta.rawQuery,
-          meta.variables,
-        );
+        const response = await _client.request<BaseRecord>({
+          document: meta.rawQuery,
+          variables: meta.variables,
+          requestHeaders: headers,
```
igordonin, line 452: Oops, those aren't needed
```suggestion
          requestHeaders: headers,
```

packages/nestjs-query/src/dataProvider/index.ts
```diff
@@ -472,7 +480,11 @@ const dataProvider = (client: GraphQLClient): Required<DataProvider> => {
             variables = gqlMutation.variables;
           }
 
-          const response = await client.request<BaseRecord>(query, variables);
+          const response = await _client.request<BaseRecord>({
+            document: query,
+            variables,
+            requestHeaders: headers,
```
igordonin, line 486: Oops, those aren't needed
```suggestion
          requestHeaders: headers,
```

packages/nestjs-query/src/dataProvider/index.ts
```diff
@@ -419,30 +419,38 @@ const dataProvider = (client: GraphQLClient): Required<DataProvider> => {
...
 
       if (gqlOperation) {
-        const response: any = await client.request(
-          gqlOperation,
-          meta?.variables ?? {},
-        );
+        const response: any = await _client.request<BaseRecord>({
+          document: gqlOperation,
+          variables: meta?.variables,
+          requestHeaders: headers,
```
igordonin, line 442: Oops, those aren't needed
```suggestion
          requestHeaders: headers,
```

BatuhanW (commented):
Hey @igordonin thanks for the PR! I'll do a more in-depth review later but, wanted to mention that we should keep existing POST request mocks and tests instead of replacing them with GET. Better to have both of them.

igordonin:
> Hey @igordonin thanks for the PR! I'll do a more in-depth review later but, wanted to mention that we should keep existing POST request mocks and tests instead of replacing them with GET. Better to have both of them.

Np, Glad to help.

Agreed. We still have both though. The ones I replaced were being used to test GET requests. I have not removed any of the previous tests doing POST requests.

BatuhanW:
> > Hey @igordonin thanks for the PR! I'll do a more in-depth review later but, wanted to mention that we should keep existing POST request mocks and tests instead of replacing them with GET. Better to have both of them.
> 
> Np, Glad to help.
> 
> Agreed. We still have both though. The ones I replaced were being used to test GET requests. I have not removed any of the previous tests doing POST requests.

I see, I was confused by the diff and poor test case naming!

We also wanted to allow users to make REST requests if they need to, for cases like custom authentication or one off requests like PATCH, PUT into endpoints. Do you think we could do that in this custom method as well?

igordonin:
> We also wanted to allow users to make REST requests if they need to, for cases like custom authentication or one off requests like PATCH, PUT into endpoints. Do you think we could do that in this custom method as well?

Not as is, no. It's currently using `gql-request` to make the requests, so it only supports `GET` and `POST`. So this whole DataProvider is currently coupled to GraphQL not only because of the dependencies but also because of the implementations.

I don't think you should try to provide a single generic DataProvider though...

BatuhanW:
Hey @igordonin looks great! I've discussed with core team and we've realized an edge-case that can be handled here.

https://github.com/refinedev/refine/blob/71ee737c2bfd89f928546d66d2f78b1b4cc1c544/examples/app-crm/src/providers/data/index.ts#L15

In `app-crm` example, we are overriding `fetch` method of gql client to add refresh token mechanism.

I think we can do something like this in custom method:

```
const _client = new GraphQLClient(url, {...client.requestConfig, headers })
```

This way, (hopefully) we can preserve user's client configuration.

igordonin:
Ok, done that. Now we keep the middlewares and override only headers and http method.

packages/nestjs-query/src/dataProvider/index.ts
```diff
@@ -419,30 +419,37 @@ const dataProvider = (client: GraphQLClient): Required<DataProvider> => {
...
+      if (!SUPPORTED_METHODS.some((it) => it === method)) {
+        throw Error(`GraphQL does not support ${method} method.`);
       }
 
+      const validMethod = method as "get" | "post";
+
+      const _client = new GraphQLClient(requestUrl, {
+        ...client.requestConfig,
+        method: validMethod,
+        headers,
```
BatuhanW, line 434:
```suggestion
        headers: { ..client.requestConfig.headers, ...headers },
```

I think would be nice to not lose headers, if there are any.

BatuhanW (commented):
Great PR! Thanks for the updates, added a minor comment. We'll release it with August release.

packages/nestjs-query/src/dataProvider/index.ts
igordonin: nice catch

BatuhanW (approved):
Thanks for the great PR! Great stuff. 🚀

PR #6185, opened 2024-07-25
feat(devtools): ability to change ports and domains in devtools server

### LINKED ISSUE #5111: [BUG] Devtools cannot be used outside of `localhost`
### Describe the bug

I would like use Devtools in my Refine project. I'm running my development in Docker containers and since Devtools bind exclusively on `localhost`, Devtools client cannot reach the server.

### Steps To Reproduce

1. Start a new Devtools project
2. Observe it's binding on `localhost` and no other network interface
3. Look for a configuration option for defining binding IP.
4. Fail to find anything and cry a little in despair :) 

### Expected behavior

Binding interface (and port) should be configurable.

### Screenshot

_No response_

### Desktop

_No response_

### Mobile

_No response_

### Additional Context

There has been some attempts to implement port configuration, which have been reverted in https://github.com/refinedev/refine/pull/5056, with a [promise](https://github.com/refinedev/refine/pull/5056#discussion_r1345475273) to provide a solution in the future.

### LINKED ISSUE #5803: [BUG] env utility breaks Vite dotenv conventions
### Describe the bug

Vite uses dotenv (and a plugin or two) to load values from files with the following conventions:

```
.env                # loaded in all cases
.env.local          # loaded in all cases, ignored by git
.env.[mode]         # only loaded in specified mode
.env.[mode].local   # only loaded in specified mode, ignored by git
```

Additionally, it behaves according to the following description:

> An env file for a specific mode (e.g. .env.production) will take higher priority than a generic one (e.g. .env).
> 
> In addition, environment variables that already exist when Vite is executed have the highest priority and will not be overwritten by .env files.

([reference](https://vitejs.dev/guide/env-and-mode))

The Refine CLI automatically loads values in the `.env` file which appear to persist in the environment when executing the Vite commands (i.e. `vite dev`).

https://github.com/refinedev/refine/blob/16eefc493da3e66a095a11d9dcbeff0ec64dca57/packages/cli/src/utils/env/index.ts#L3

These take precedence and effectively break the `.env.local`, `.env.development`, etc. conventions of Vite.

CRA has slightly different but similar conventions. I assume the Refine CLI would be affecting it similarly?

### Steps To Reproduce

1. Create a Refine/Vite project.
2. Create a `.env` file with the content `VITE_TITLE=production`.
3. Create a `.env.development` file with the content `VITE_TITLE=development`.
4. Run `npm run refine dev`.
5. Observe the value of `import.meta.env.VITE_TITLE` is "production", not "development".

### Expected behavior

The Refine CLI shouldn't break Vite's dotenv conventions. In the trivial example provided, the value of `import.meta.env.VITE_TITLE` should be "development" from the `.env.development` file.

### Packages

- "@refinedev/cli": "2.16.27"
- "@refinedev/core": "4.48.0"
- "@refinedev/kbar": "1.3.7"
- "@refinedev/mui": "5.14.5"
- "@refinedev/react-hook-form": "4.8.15"
- "@refinedev/react-router-v6": "4.5.6"
- "vite": "5.2.7"
- package manager: Yarn 4.1.1 w/ node_modules linker

### Additional Context

I explicitly renamed my `.env` file to `.env.production` to "workaround" the issue. That file appears to not be loaded by the Refine CLI, and then my app builds as expected. It's an unfortunate workaround, effectively forcing us to not use the `.env` file and "layering" of settings.

aliemir:
## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

## Changes

fix(cli): avoid polluting `process.env` with unwanted environment variables

Previously, the `@refinedev/cli` used `dotenv` to load environment variables from `.env` files and populate `process.env`. This caused issues when the users app has a different convention for environment variables, e.g. `.env.development`, `.env.production`, etc.

Now, the `@refinedev/cli` will read the file but avoid populating `process.env` with the variables and keep the values in its scope without passing them to the child processes. This will prevent unwanted environment variables from being passed to the child processes and avoid conflicts with the user's environment variables.

Resolves #5803

feat(devtools): ability to change the port of the devtools server

Now users can change the port of the devtools server by setting the `REFINE_DEVTOOLS_PORT` environment variable. Previously, the port was hardcoded to "5001" and could not be changed.

If you're using `@refinedev/cli`'s runner commands to start your development server, `REFINE_DEVTOOLS_PORT` will be propagated to your app with appropriate prefix. E.g. if you're using Vite, the environment variable will be `VITE_REFINE_DEVTOOLS_PORT` and it will be used by the `@refinedev/devtools`'s `<DevtoolsProvider />` component to connect to the devtools server.

- In Next.js apps, it will be prefixed with `NEXT_PUBLIC_`
- In Craco and Create React App apps, it will be prefixed with `REACT_APP_`
- In Remix apps and other custom setups, the environment variable will be used as is.

In some scenarios where the environment variables are not passed to the browser, you may need to manually set the Refine Devtools URL in the `<DevtoolsProvider />` component via the `url` prop. Remix apps do not automatically pass environment variables to the browser, so you will need to set the URL manually. If not set, the default URL will be used.

While the port can be changed, this feature also allows users to host the devtools server on a different machine or domain and provide the `<DevtoolsProvider />` with the custom domain URL. This such case will be useful if you're dockerizing your app and devtools server separately.

**Enterprise Edition**: Refine Devtools running on ports other than "5001" is only available in the Enterprise Edition. If you're using the Community Edition, Refine Devtools will not work if the port is changed.

Resolves #5111


BatuhanW (changes requested):

documentation/docs/enterprise-edition/devtools/docker.tsx
```diff
@@ -0,0 +1,203 @@
...
+RUN npm install
+
+# Copy the rest of the application
+COPY . .
+
+# Expose the port the app runs on
+EXPOSE 5173
+
+# Command to run the development server
+CMD ["npm", "run", "docker"]
```
BatuhanW, line 107:
```suggestion
CMD ["npm", "run", "dev"]
```

documentation/docs/enterprise-edition/devtools/docker.tsx
```diff
@@ -0,0 +1,203 @@
...
+`.trim();
+
+const DockerfileDevtoolsCode = /* dockerfile */ `
+# We're setting up our Devtools server and running it on port 5001.
+# We'll then use Nginx to reverse proxy the requests to the correct services.
+# We're running devtools at port 5001 and we'll access it via http://devtools.local.
+
+# Use the official Node.js image as a parent image
```
BatuhanW, line 115:
```suggestion
# Use the Refine's Node.js image as a parent image
```

BatuhanW (approved):
Great stuff. ❤️

Issue #5864, opened 2024-04-18
[FEAT] Passing arbitrary graphql variables to hasura dataProvider


tomoemon:
### Is your feature request related to a problem? Please describe.

I am very pleased that gqlQuery allows us to perform arbitrary GraphQL Queries against hasura.
https://github.com/refinedev/refine/issues/5489

However, it is not possible to use hasura's powerful filters because only CrudFilter[] can be passed to filters in the useTable. It is also not possible to use variables type information generated by graphql codegen.

In my case.
Given a type with the relation User has many Books, I can write a query in hasura to get a list of Users who have 3 or more Books, which is not possible in the current CrudFilter.

### Describe alternatives you've considered

added to the following line in hasura dataProvider.
```typescript
        variables = {
          ...hasuraPagination,
          ...(hasuraSorting &&
            (namingConvention === "graphql-default"
              ? {
                orderBy: hasuraSorting,
              }
              : {
                order_by: hasuraSorting,
              })),
          ...(hasuraFilters && {
            where: hasuraFilters,
          }),
          ...(meta?.gqlVariables && meta?.gqlVariables), // added
        };
```
https://github.com/refinedev/refine/blob/master/packages/hasura/src/dataProvider/index.ts#L181

This allows the addition of any where condition supported by hasura, using type information hints, as follows.
```typescript
  const { tableProps } = useTable<GetFieldsFromList<BlogPostsQuery>>({
    syncWithLocation: true,
    meta: {
      gqlQuery: BlogPostsDocument,
      gqlVariables: {
        where: {
          _and: [
            {
              title: {
                _ilike: "%test%",
              },
            }
          ],
        },
      } as BlogPostsQueryVariables, // generated type from graphql query
    },
  });
```

### Describe the thing to improve

- add `gqlVariables?: any` to `GraphQLQueryOptions`
  - Ideally, I would like to pass the type information generated from the graphql query by type parameter instead of any, but I can't think of a way to achieve that right now.
- use `meta.gqlVariables` in `getList`, `getMany` of hasura dataProvider

alicanerdurmaz:
Hello @tomoemon, thanks for the issue. 

It seems like a good idea, but can you give me the GQL Query you want to generate so I can better understand the problem? Maybe we can make CrudFilter[] to support this.

tomoemon:
@alicanerdurmaz Thank you for your response!

For example, I want to execute the following query

Trying to get a list of categories that have at least one Posts with the string "test" in the title.
[...]

PR #6222, opened 2024-08-01
feat(hasura): allow passing arbitrary GraphQL variables using meta.gqlVariables

Sergio16T:

## Bugs / Features

- [x] Related issue(s) linked

Current Hasura Data Provider does not support additional `meta.gqlVariables` to support more advanced hasura filtering

## What is the new behavior?

Core Package GraphQLQueryOptions has additional field `gqlVariables` to support more advanced filters used in data providers. 

Hasura Package Data Provider `getList` and `getMany` have been updated to support merging `meta.gqlVariables.where` into other provided hasura filters.  Other `meta.gqlVariables` are added as well. 
 
These filters are applied to each query and are currently used to reduce the dataset in table. 
**Warrants further discussion:** Please see Notes for reviewers.
1. Any subsequent changes to table filters will not affect `meta.gqlVariables`. 
2. If `meta.gqlVariables` contain table column values these values are not included in table column filter dropdown as selected filters. 
3. If Table Column Filter Clear's current filters. These will not affect the `meta.gqlVariables` filters used on each subsequent query. 

feat (#5864 )

**Note for reviewers:** 
This PR is still in Draft and needs further discussion. See below about current work that was considered out of scope for first draft. 

**Out of scope:**

1. This PR needs further discussion about initial filters and meta.gqlVariables. See below. 
3. I did not makes changes to nest-js or graphql packages. I wanted to wait to implement changes to these subsequent packages until we discussed desired behavior with changes. 
4. Changes to documentation. (Can be done in subsequent commits). 

Current Implementation in first commit: 

`meta.gqlVariables.where` contains hasura filters that are used to reduce the data set returned by the graphQL query and any table column filters will apply additional filters to query. `meta.gqlVariables` can include any gql variables used by query. 

1. Any additional filtering done in table will apply additional filters on the dataset returned by the gqlQuery with gqlVariables applied. 
2. If the `gqlVariables` happen to contain table column fields these fields are not included in initial filter selection displayed or affected by later clearing of fields in table column filtering. 

Is this desired behavior for **gqlVariables to reduce the initial dataset. And then any subsequent filtering done by table only applies additional filters to this reduced dataset?** 

OR would we like to increase the scope of this PR to update `useTable` Hook to apply `meta.gqlVariables` to the initial filters used by table. Then any subsequent clearing of these filters would also need to change the subsequent queries? I believe this is possible by adding logic to check for fields in `meta.gqlVariables` that are contained in table in `useTable` hook.


Sergio16T:
Sounds good @aliemir! Thank you. Congratulations on August Release 🚀🚀

packages/hasura/src/utils/boolexp.ts
BatuhanW, line 1: Great addition!

BatuhanW: Let's add this file to `packages/hasura/refine.config.js`. It's used by CLI when users swizzle the data provider.

packages/hasura/src/dataProvider/index.ts
```diff
@@ -159,7 +173,11 @@ const dataProvider = (
         ? generateSorting(sorters)
         : upperCaseValues(camelizeKeys(generateSorting(sorters)));
 
-      const hasuraFilters = generateFilters(filters, namingConvention);
+      let hasuraFilters = generateFilters(filters, namingConvention);
+      hasuraFilters = mergeHasuraFilters(
+        hasuraFilters,
+        meta?.gqlVariables?.where,
+      );
```
BatuhanW, line 180: We can move it to lower part, where we check `meta?.gqlQuery`. I think this wouldn't really work if user isn't providing it's own gql query.

packages/hasura/src/utils/generateFilters.ts
```diff
@@ -172,3 +174,29 @@ export const generateFilters: any = (
 
   return nestedQuery;
 };
+
+export const mergeHasuraFilters = (
```
BatuhanW, line 178: Do we need to check for naming convention here?

BatuhanW: Let's make sure to add additional specs for this one.

packages/hasura/src/dataProvider/index.ts
BatuhanW: We can also add additional test cases for this one.

BatuhanW (commented):
Hey @Sergio16T thanks for the PR with great description! PR Looks good generally, missing some small things here and there.

> gqlVariables to reduce the initial dataset. And then any subsequent filtering done by table only applies additional filters to this reduced dataset?

Yes. I think that would be really hard, if not impossible to sync table with custom filters provided by users. That would increase complexity. At least for now, let's keep it simple and just support additional variables, and then we can see better if and further updates are needed.

Also let's make sure to add tests covering the changes. We are using nock to mock requests. To record requests for new tests, you can uncomment line 4 at `packages/hasura/test/jest.setup.ts`. When you run the test, it should make the request, so you can see if it works as expected or not. Thenyou wouldn't need to enable it for subsequent tests, once you get real response.

packages/hasura/src/utils/generateFilters.ts
Sergio16T: @BatuhanW, @aliemir  I did a deep dive into Hasura Naming Conventions. Learned quite a bit about Hasura Filtering during the process

I've written logic to convert nested filters from gqlVariables to `graphql-default` casing. I'm continuing to add test cases with different filters and reviewing. 

1. This change makes it so that if a dev writes gqlVariables.where filters with the`hasura-default` casing then the data provider will convert the variables to `graphql-default` camelCase when the naming convention is `graphql-default`. 

That said I wonder if we should avoid this conversion entirely and encourage developer to follow the naming convention that is defined in their application when writing these advanced hasura filters. 

1. This conversion increases complexity. 
2. This conversation required several exhaustive checks.
3. Would reduce maintenance. 

Since these filters are written in the hasura filter query syntax is there any benefit to us doing this conversion out of the box? 

1. By doing this conversion natively developers can opt for which ever query syntax is most comfortable for them and the data provider will do the conversion. 
2. Without this conversion we get API error's if the dev uses incorrect syntax e.g. `created_at` or `is_null` when naming convention is `graphql-default` e.g. `error: field 'created_at' not found in type: 'PostsBoolExp ...` Then the dev will need to make updates to the `meta.gqlVariables` to use supported fields and operators e.g. `createdAt` `isNull` based on the graphQL schema and naming convention in use. 

After doing the dev work and learning about naming conventions and hasura query syntax I'm open to either solution. It was a valuable learning experience. 

What do you guys think?

BatuhanW: > My instinct is that it would make the most sense to encourage developers to use the Hasura Query Syntax based on the selected naming convention instead of doing the conversion. I believe we don't need to do any name convention conversion logic for the gqlVariables and I can remove the convertNestedFiltersToGraphqlDefault logic in a subsequent commit.

Yeah, you are right. We shouldn't do some magic under the hood. Since it's provided by users, they'll already use the right naming convention.

BatuhanW (approved):
Great work! @Sergio16T  Thanks for the PR 🚀

aliemir (approved):

PR #6246, opened 2024-08-11
feat(cli): add Remix SPA mode support

### LINKED ISSUE #6127: [FEAT] Add Script for Remix SPA Mode

In Remix SPA Mode, the scripts differ from those in the regular Remix setup.

```json
"dev": "remix vite:dev",
"build": "remix vite:build",
"start": "vite preview",
```

Currently, the refine CLI only supports Remix SSR Mode.

Using [npm-run-all](https://www.npmjs.com/package/npm-run-all).

Example: `npm run dev`
```json
"devtools": "refine devtools",
"dev:remix": "remix vite:dev",
"dev": "npm-run-all --parallel devtools dev:remix",
```

### Additional context

_No response_

Whether SPA Mode is enabled can be determined from `vite.config.ts`. While it would be ideal if `getProjectType` could detect this, it seems more reasonable to make it selectable through the platform option, as shown below:

`refine dev -p remix-spa`

YusukeSano:
## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

Scripts for Remix SPA mode are not supported.

## What is the new behavior?

It is now possible to execute the Remix SPA Mode script by selecting it from the platform options.
resolves #6127

### DIFF
--- a/packages/cli/src/commands/runner/projectScripts.ts
+++ b/packages/cli/src/commands/runner/projectScripts.ts
@@ -50,6 +50,15 @@ export const projectScripts = {
       return require.resolve(`.bin/${binName}`);
     },
   },
+  [ProjectTypes.REMIX_SPA]: {
+    getDev: (args: string[]) => ["vite:dev", ...args],
+    getStart: (args: string[]) => ["preview", ...args],
+    getBuild: (args: string[]) => ["vite:build", ...args],
+    getBin: (type?: "dev" | "start" | "build") => {
+      const binName = type === "start" ? "vite" : "remix";
+      return require.resolve(`.bin/${binName}`);
+    },
+  },
   [ProjectTypes.CRACO]: {
     getDev: (args: string[]) => ["start", ...args],
     getStart: (args: string[]) => ["start", ...args],
--- a/packages/cli/src/commands/runner/projectScripts.ts
+++ b/packages/cli/src/commands/runner/projectScripts.ts
@@ -50,6 +50,34 @@ export const projectScripts = {
       return require.resolve(`.bin/${binName}`);
     },
   },
+  [ProjectTypes.REMIX_VITE]: {
+    getDev: (args: string[]) => ["vite:dev", ...args],
+    getStart: (args: string[]) => {
+      // remix-serve accepts a path to the entry file as an argument
+      // if we have arguments, we will pass them to remix-serve and do nothing.
+      // ex: `refine start ./build/server/index.js`
+      const hasArgs = args?.length;
+      if (hasArgs) {
+        return args;
+      }
+
+      // otherwise print a warning and use `./build/server/index.js` as default
+      console.log();
+      console.warn(
+        "🚨 Remix requires a path to the entry file. Please provide it as an argument to `refine start` command in package.json scripts",
+      );
+      console.warn("Refine will use `./build/server/index.js` as default");
+      console.warn("Example: `refine start ./build/server/index.js`");
+      console.log();
+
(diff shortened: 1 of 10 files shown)


packages/cli/src/commands/runner/projectScripts.ts
```diff
@@ -50,6 +50,15 @@ export const projectScripts = {
       return require.resolve(`.bin/${binName}`);
     },
   },
+  [ProjectTypes.REMIX_SPA]: {
```
alicanerdurmaz, line 53: CLI uses them in the `start` or `dev` [commands.](https://github.com/refinedev/refine/blob/e777c467f32c167d6a2ab13a9bb481f6abcbac2d/packages/cli/src/commands/runner/dev/index.ts#L43) But you didn't make any changes on [`getProjectType`](https://github.com/refinedev/refine/blob/e777c467f32c167d6a2ab13a9bb481f6abcbac2d/packages/cli/src/utils/project/index.ts#L4) function. Is this working according to your needs?

alicanerdurmaz (changes requested):

packages/cli/src/commands/runner/projectScripts.ts
alicanerdurmaz: I guess it's working works when we specify the project like this `refine dev -p remix-spa`. It would be better if the CLI automatically detected that. What do you think @BatuhanW @aliemir ?

YusukeSano: Thank you for your feedback!

Do you have any suggestions for a good approach to resolving the `vite.config` file?
If automatic detection adds significant complexity, I would prefer to create a separate issue for that enhancement rather than including it in this PR.

Looking forward to your thoughts!

aliemir (changes requested):
Hey @YusukeSano thank you for the PR! I think we can do couple of small changes;

- Let's have two new project types `REMIX_VITE` and `REMIX_SPA` (or `REMIX_VITE_SPA`). Probably only difference will be the `start` command and its executable. 

- In the `getProjectType` method, we can check for `vite` and `remix` dependencies. Like we do in rest of the types. No need for checking `vite.config.ts` etc.

- `REMIX_VITE` can use the configuration you added in the PR for the `dev` and `build` commands but Remix + Vite without SPA mode should use `remix-serve` for the `start` command. You can check the `getBin` and `getStart` properties of the current Remix type and see how we implemented before.

- `REMIX_VITE_SPA` should use `vite preview` command for the start command and therefore `getStart` and `getBin` should be updated in according to these. Your current changes for the `getStart` and `getBin` can work there.

- `getProjectType` method shouldn't check for the SPA mode and only check for `REMIX_VITE` based on the dependencies of the project.

- If a user is using Remix + Vite, we'll be able to successfully determine the `dev` and `build` commands. Only difference will be the `start` command between Remix + vite and spa mode. For SPA mode users, they can use `refine start -p remix-vite-spa` and explicitly pass the project type, then we'll use the `vite preview` for it.

Let me know if this is clear and if you can work on this, we'll be happy to review the changes 🙏

packages/cli/src/commands/runner/projectScripts.ts
aliemir: Left a separate comment also clearing the issue of this conversation. Please check @alicanerdurmaz and @YusukeSano 🙏

aliemir (approved):
Hey @YusukeSano thank you for your contribution and patience. Great job 🚀 🚀 This will be included in our next release 🙏

alicanerdurmaz (approved):

PR #6386, opened 2024-10-03
fix(core): wrap `setFiltersFn` with `useCallback`

### LINKED ISSUE #6385: [BUG] setFilters from useDataGrid causes infinite loop inside useEffects
### Describe the bug

I was trying to use the `setFilters` from the useDataGrid hook in a custom debounce function that are attached with a state. I have something very simple like:
```
useEffect(() => {
    setFilters(
      [
        {
          field: "id",
          value: debouncedSearchInput,
          operator: "eq",
        },
      ],
      "replace"
    );
  }, [debouncedSearchInput, search,  setFilters]);
```

The code above is causing a infinite loop. After investigate a bit I realised that inside the `useTable` hook the `setFiltersFn` are not wrapped in a `useCallback` and this causing changes in states changes the reference of the function and bc of that it's impossible pass the `setFilters` as a dependency.

The fix is quite easy, I can open a PR for fixing it.

### Steps To Reproduce

1. get the `setFilters` from `useDataGrid`
2. use it in a useEffect and pass setFilters as a dependency.
3. infinite loop

### Expected behavior

Use setFilters as a dependency should not cause infinite loop

### Packages

- @refinedev/core

### Additional Context

_No response_

hugorezende:
## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

Adding the `setFilters` as a dependency in the useEffect hook causes infinite loop since setFiltersFn are not wrapped in `useCallbacks`

## What is the new behavior?

No infinite loop

fixes #6385

### DIFF
--- a/.changeset/spotty-jokes-swim.md
+++ b/.changeset/spotty-jokes-swim.md
@@ -1,0 +1,9 @@
+---
+"@refinedev/core": patch
+---
+
+fix: fix setFiltersFn #6385
+
+With this we can use the setFilters as dependencies inside useEffects without infinite loop since state changes in the hook won't cause the functions to be re-assigned
+
+[Fixes #6385](https://github.com/refinedev/refine/issues/6385)
--- a/.changeset/spotty-jokes-swim.md
+++ b/.changeset/spotty-jokes-swim.md
@@ -2,7 +2,7 @@
 "@refinedev/core": patch
 ---
 
-fix: fix setFiltersFn #6385
+fix(core): wrap `setFilters` and `setSorters` methods with `useCallback` to prevent looping re-renders
 
 With this we can use the setFilters as dependencies inside useEffects without infinite loop since state changes in the hook won't cause the functions to be re-assigned
 
--- a/packages/core/src/hooks/useTable/index.ts
+++ b/packages/core/src/hooks/useTable/index.ts
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from "react";
+import React, { useState, useEffect, useCallback } from "react";
 
 import type {
   QueryObserverResult,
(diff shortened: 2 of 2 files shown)


packages/core/src/hooks/useTable/index.ts
```diff
@@ -520,36 +520,49 @@ export function useTable<
...
-      unionFilters(preferredPermanentFilters, newFilters, prevFilters),
-    );
-  };
+  const setFiltersAsMerge = useCallback(
+    (newFilters: CrudFilter[]) => {
+      setFilters((prevFilters) =>
+        unionFilters(preferredPermanentFilters, newFilters, prevFilters),
+      );
+    },
+    [setFilters, unionFilters, preferredPermanentFilters],
```
aliemir, line 529: `setFilters` is a SetStateAction and `unionFilters` is an imported util, they don't need to be included in the dependency arrays.

aliemir: Also can you make a similar update in the `setSortWithUnion` function as well? 🙏

.changeset/spotty-jokes-swim.md
```diff
@@ -0,0 +1,9 @@
+---
+"@refinedev/core": patch
+---
+
+fix: fix setFiltersFn #6385
```
aliemir, line 5:
```suggestion
fix(core): wrap `setFilters` and `setSorters` methods with `useCallback` to prevent looping re-renders
```

Having a bit more info in the first line will be great for the changelogs 😅

aliemir (changes requested):
Hey @hugorezende thank you for the PR! I've made a small change request in the changeset and left two comments for the dependency arrays, can you check them and apply the necessary changes? 🙏 

Then, we can include this in our next release 🚀 🚀

packages/core/src/hooks/useTable/index.ts
hugorezende: hey @aliemir, thanks for checking the PR. I'm just so used with the exhaustive rule i didn't realised it 😄  
Just fixed those and also added it on `setSortWithUnion` 
990f8e9

alicanerdurmaz (approved):

PR #6503, opened 2024-11-18
fix(mui): fixed condition preventing Breadcrumb text display

### LINKED ISSUE #6497: [FEAT] Breadcumbs configure displayed elements
### Describe the bug

It appears that the breadcrumb component doesn't display anything when the array contains only a single element. This undermines the consistency of the layout, as I would like to maintain a consistent breadcrumb style across all pages, even when there is only one breadcrumb item. Could you kindly add a property to define the minimum number of elements required for the breadcrumb to be displayed?

### Steps To Reproduce

```js
import { Toolbar } from "@mui/material";
import { useBreadcrumb } from "@refinedev/core";
import { Breadcrumb, ThemedLayoutV2 } from "@refinedev/mui";
import { Outlet } from "react-router-dom";

const Header = () => {
  const breadcumb = useBreadcrumb();
  console.log(breadcumb);

  return (
    <Toolbar>
      <Breadcrumb showHome meta={{ projectId: "123" }} key={1} />
    </Toolbar>
  );
};

const SidebarLayout = () => (
  <ThemedLayoutV2 Header={Header}>
    <Outlet context={{ component: "main", sx: { flex: 1 } }} />
  </ThemedLayoutV2>
);

export default SidebarLayout;
```

### Expected behavior

See above.

### Packages

Latest.

### Additional Context

_No response_

aress31:
## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

## What is the new behavior?

fixes #6497 

## Notes for reviewers

The following tasks have not been performed, kindly complete them to provide me with a how-to example:


BatuhanW (changes requested):
Hey @aress31 thanks for taking time to create issues and the PR.

In your PR, please do not include any irrelevant changes. Only the changes related to your fix and specs. 

Imports like `import HomeOutlined from "@mui/icons-material/HomeOutlined";` are not coincidence. They are intentionally like that because of bundling issues with mui library.

BatuhanW:
@aress31 For tests, you can update the tests in `packages/mui/src/components/breadcrumb/index.spec.tsx` file. We have `routerInitialEntries: ["/posts/create"],` there, you can add new cases with `routerInitialEntries: ["/posts"],` and see if it renders. You can also stash your changes and run the test to see if new test correctly fails.

aress31:
> Hey @aress31 thanks for taking time to create issues and the PR.
> 
> In your PR, please do not include any irrelevant changes. Only the changes related to your fix and specs.
> 
> Imports like `import HomeOutlined from "@mui/icons-material/HomeOutlined";` are not coincidence. They are intentionally like that because of bundling issues with mui library.

@BatuhanW besides reverting the imports, what else do you want me to change? I think refactoring the file as I did help with modularity and readability - killing two birds with one stone.

aress31:
As requested, I focused solely on the fix. However, I recommend reviewing my previous commits with the refactoring. Some styles, such as default padding and `fontSize`, should, in my opinion, be removed to keep the component unstyled, allowing users to customize it via `breadcrumbProps`. Additionally, I suggest improving modularity by splitting the breadcrumb and home rendering logic.

alicanerdurmaz (changes requested):
Hello @aress31, Thanks for improving Refine 🚀

I added comments to the issue [about](https://github.com/refinedev/refine/issues/6497#issuecomment-2488156514) implementation. Please feel free to ask questions if you have any concerns

packages/antd/src/components/breadcrumb/index.tsx
```diff
@@ -38,9 +39,7 @@ export const Breadcrumb: React.FC<BreadcrumbProps> = ({
 
   const ActiveLink = routerType === "legacy" ? LegacyLink : Link;
 
-  if (breadcrumbs.length === 1) {
-    return null;
-  }
+  if (breadcrumbs.length >= minItems) return null;
```
alicanerdurmaz, line 42: I believe it should be like this:

```suggestion
  if (breadcrumbs.length < minItems) return null;
```

alicanerdurmaz (changes requested):

packages/antd/src/components/breadcrumb/index.tsx
```diff
@@ -38,9 +39,7 @@ export const Breadcrumb: React.FC<BreadcrumbProps> = ({
 
   const ActiveLink = routerType === "legacy" ? LegacyLink : Link;
 
-  if (breadcrumbs.length === 1) {
-    return null;
-  }
+  if (breadcrumbs.length < minItems) return null;
```
alicanerdurmaz, line 42: - case 1: breadcrumbs.length = 0:  // both versions return null
- case 2: breadcrumbs.length = 1:  // old version returns null, new version shows
- case 3: breadcrumbs.length = 2:  // Both versions renders 

we have two options to fix case 2:
1. 
```suggestion
  if (breadcrumbs.length <= minItems) return null;
```
2. default should be set to 2 `minItems = 2`

I believe second option is better because it fits `minItems` name.

alicanerdurmaz (changes requested):

packages/antd/src/components/breadcrumb/index.tsx
aress31: I agree with point 1 but why should the default be set to 2? What is the rational behind this? If there is a configured route, the breadcrumb should show it be default unless instructed otherwise? 2 also seems arbitrary, why not3? :)

alicanerdurmaz: Because previously, the breadcrumb would render only if there were at least 2 items, but it wouldn’t render if there was only 1 item. So, `minItems` should be 2. If we break this logic, it would be a breaking change, and current users who update Refine would see a breadcrumb even in cases with just 1 item.

I'm with you. If the route is configured, it should render, but we shouldn't make breaking changes, even if the old implementation is wrong.

aress31: We should be good to merge then since I have the default for `minItems` to be `2`. Please let me know if there is another blocker.

Also for future major release might be a good idea to implement the breaking change of setting `minItems` to `1`.

alicanerdurmaz: Everything is good, thanks but `should not render breadcrumb when the number of items is less than minItems` This test is not passing

aress31: @alicanerdurmaz should be fixed now, kindly review.

Issue #6651, opened 2025-01-17
[BUG] [Supabase] When a "in" filter is nested it is parsed incorrectly


jkronlachner:
### Describe the bug

If a `LogicalFilter` with the operator `in` is used inside a `ConditionalFilter` `or` the array value is parsed incorrectly by the `generateFilter` Method. 

When passing a array outside of a `LogicalFilter` it parses correctly. 

### Steps To Reproduce

I am using the following Filter configuration: 
```javascript
{
        operator: 'or',
        value: [
        {
           field: "id",
           operator: 'in',
           value: [1,2,3],
        },
        ... more filters
    ],
},

```
This is sending the following to supabase as filter: 
`http://127.0.0.1:54321/rest/v1/company?select=*&offset=0&limit=30&or=(id.in."1","2","4")`

Supabase then throws the follwing error: 

```
{
    "code": "PGRST100",
    "details": "unexpected \"1\" expecting \"(\"",
    "hint": null,
    "message": "\"failed to parse logic tree ((id.in.1,2,4))\" (line 1, column 10)"
}
```

### Expected behavior

This is missing the parentheses around the array. It should look like this:
`http://127.0.0.1:54321/rest/v1/company?select=*&offset=0&limit=30&or=(id.in.("1","2","4"))
`
### Packages

 - @refinedev/supabase

PR #6652, opened 2025-01-17
fix(supabase): fixed 'in'-filter when used inside 'or'

jkronlachner:
If a `in` Filter is used inside a `LogicalFilter` it is not parsed correctly to supabase. See #6651 

## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

Filter get parsed incorrectly `or=(id.in.1,2,4)`

## What is the new behavior?
Filter get parsed correctly `or=(id.in.("1", "2"))`

fixes (#6651)


packages/supabase/test/getList/index.mock.ts
```diff
@@ -2073,3 +2073,77 @@ nock("https://iwdfzvfqbtokqetmbmbp.supabase.co:443", {
       'h3=":443"; ma=86400',
     ],
   );
+
+nock("https://iwdfzvfqbtokqetmbmbp.supabase.co:443", {
```
BatuhanW, line 2077: Let's please record the actualy request. Here we are providing the mock data, without making the request, and we are testing it based on the data provided by us. It's not really testing anything.

BatuhanW (commented):
Thanks for the PR, added a comment about nock records.

packages/supabase/test/getList/index.mock.ts
jkronlachner: yeah, already thought so... is there a reason why the other tests also don't make actual requests or am i reading something wrong? 

nevertheless will change to a real request

BatuhanW: They are recorded from actual requests, so next time you do the same request, nock will return recorded response. You can check jest.setup.ts file, there should be a commented line goes like `nock.recorder.rec()`, you can uncomment it and run the single test you need, it should print request/response to the console. Then you can paste it here.

BatuhanW:
@jkronlachner are you gonna be able to continue this PR?

jkronlachner:
@BatuhanW yes, should be done by the end of this week

jkronlachner:
@BatuhanW i think i got it, now the test is the response from the nock.recorder.rec

BatuhanW (approved):

PR #6757, opened 2025-04-21
feat(ShowComponent): add meta prop for button visibility control

write-arsalan-nawaz:
This PR adds support for using the `meta` prop in the `<Show>` component (both Mantine and MUI) to control the visibility of Edit and Delete buttons.

### What's Added
- `meta.canDelete` and `meta.canEdit` support
- Fallback preserved to `canDelete` and `canEdit` props

Closes #7951

### DIFF
--- a/packages/antd/src/components/crud/show/index.tsx
+++ b/packages/antd/src/components/crud/show/index.tsx
@@ -10,6 +10,7 @@ import {
   useRouterType,
   useBack,
   useGo,
+  useShow,
 } from "@refinedev/core";
 
 import {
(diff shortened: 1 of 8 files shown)


alicanerdurmaz:
Hi @write-arsalan-nawaz,

Thanks for the PR! I have a couple of questions:

- In `antd` you mount `useShow`, but you don’t use its return value. The other UI packages don’t mount it at all. Could you share the reason?
- You use `meta.canDelete`, yet we already have the `canDelete` prop (same for `canEdit`). Why do we need both?

I’m not sure what problem this PR fixes or which new feature it adds. Could you open an issue to explain the goal? I’ll be happy to review again after that.

For now, I’ll close the PR.

PR #6768, opened 2025-04-30
fix-types

vladbelozertsev:
I try to create custom data provider  (getList)

When i try to map filters -> ```filteres.map(f => ...)```
i cant get filter ```field``` value because of ts error: "Property field does not exist on type ConditionalFilter".

```ts
      getList: async (params) => {
        // TS error: "Property field does not exist on type ConditionalFilter"
        const filter = params.filters?.map((f) => [f.field, operators[f.operator], f.value]);
      },
```

## What is the new behavior?

No ts error

### DIFF
--- a/packages/core/src/contexts/data/types.ts
+++ b/packages/core/src/contexts/data/types.ts
@@ -320,6 +320,7 @@ export type CrudOperators =
 export type SortOrder = "desc" | "asc" | null;
 
 export type LogicalFilter = {
+  key: undefined;
   field: string;
   operator: Exclude<CrudOperators, "or" | "and">;
   value: any;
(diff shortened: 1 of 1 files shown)


alicanerdurmaz:
Hello @vladbelozertsev,

`ConditionalFilter` does not include a `field` property, this is intentional. Why do you need it? Could you share your use case?

You can distinguish between conditional and logical `CrudFilter` like this:

```ts
if ("field" in filter) {
    // TODO: handle logical filters
} else {
    // TODO: handle conditional filters
}
```

vladbelozertsev:
hi @alicanerdurmaz

i think you are right

i close this pr :)

Issue #6833, opened 2025-06-08
[BUG] useDataGrid + GraphQL: update does not receive meta.gqlMutation for editable cell


kdevo:
### Describe the bug

When using `useDataGrid` with `updateMutationOptions` and `editable: true`, the receiving hook does not get passed the metadata for determining the required GraphQL mutation when a cell/row update is performed.

### Steps To Reproduce

Data Grid:
```typescript
const { 
      dataGridProps,
     } = useDataGrid<IPost>({
        resource: "posts",
        meta: { 
            operation: "PostList",
            gqlQuery: POST_LIST, 
            gqlMutation: POST_UPDATE,
        },
        pagination: {mode: "client", pageSize:20},
        editable: true,
        updateMutationOptions: {
            meta:{
                gqlMutation: POST_UPDATE,
            },
            retry: 1,
        }
    });
```

With at least one editable column:
```typescript
            // ...
            {
                field: "name",
                headerName: "Name",
                type: "string",
                minWidth: 500,
                sortable: true,
                filterable: false,
                editable: true,
            }
```

### Expected behavior

Editable performs a call to `update` of the (GraphQL) data provider with `POST_UPDATE` as gqlOperation:

```typescript
      const { meta } = params;
      const gqlOperation = meta?.gqlMutation ?? meta?.gqlQuery;
```

### Packages

```
        "@refinedev/cli": "^2.16.21",
        "@refinedev/core": "^4.57.9",
        "@refinedev/devtools": "^1.1.32",
        "@refinedev/graphql": "^7.1.3",
        "@refinedev/inferencer": "^5.0.3",
        "@refinedev/kbar": "^1.3.6",
        "@refinedev/mui": "^6.0.0",
        "@refinedev/react-hook-form": "^4.8.14",
        "@refinedev/react-router": "^1.0.0",
```

PR #6836, opened 2025-06-10
fix(antd): Prevent unsaved dialog on modal form submission

zackseal89:
Addresses issue #6833.

When `warnWhenUnsavedChanges` is true and a form within a modal (using `useModalForm`) is dirty, submitting the form could incorrectly trigger the "unsaved changes" confirmation dialog.

This was due to a race condition where the modal closing/unmounting process initiated navigation/unload checks before React had fully processed the `warnWhen` state update to `false` (which is set synchronously in the core `useForm`'s `onFinish` handler).

The fix involves modifying the `onFinish` handler within `packages/antd/src/hooks/form/useModalForm/useModalForm.ts`. The modal's `close()` call is now wrapped in `setTimeout(..., 0)` to defer its execution to the next event loop cycle. This allows React sufficient time to process the `setWarnWhen(false)` state update, ensuring that `warnWhen` is correctly `false` when the modal close/unmount events trigger any navigation checks.

A new unit test has been added to `useModalForm`'s spec file to verify this fix and prevent regressions.

Manual end-to-end testing of this change in an example application was unfortunately blocked by unrelated build failures in `@refinedev/core` and `@refinedev/antd` packages, and subsequent tooling instability related to pre-commit hooks. The fix is verified at the unit test level.

## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

## What is the new behavior?

fixes (issue)

### DIFF
--- a/packages/antd/package-lock.json
+++ b/packages/antd/package-lock.json
@@ -1,0 +1,12548 @@
+{
+  "name": "@refinedev/antd",
+  "version": "5.46.1",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "@refinedev/antd",
+      "version": "5.46.1",
+      "license": "MIT",
+      "dependencies": {
+        "@ant-design/icons": "^5.5.1",
+        "@ant-design/pro-layout": "^7.21.1",
+        "@refinedev/ui-types": "^1.24.2",
+        "@tanstack/react-query": "^4.10.1",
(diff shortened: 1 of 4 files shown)


alicanerdurmaz:
The code and explanation don’t match the issue, and I couldn’t reproduce the bug either.

I’m closing the PR for now. We’d really appreciate it if you could open a new issue with a reproducible example.

PR #6851, opened 2025-06-23
fix: deprecated bordered prop in the Card component

damla:
## PR Checklist

## Bugs / Features

- [ ] Related issue(s) linked

image

Currently, Antdesign's Card component is giving a warning because `bordered` is deprecated. 

## What is the new behavior?

By using  `variant` instead of `bordered`, the warning is fixed.

fixes (issue)

### DIFF
--- a/packages/antd/src/components/crud/create/index.tsx
+++ b/packages/antd/src/components/crud/create/index.tsx
@@ -115,7 +115,7 @@ export const Create: React.FC<CreateProps> = ({
       >
         <Spin spinning={isLoading}>
           <Card
-            bordered={false}
+            variant="borderless"
             actions={[
               <Space
                 key="action-buttons"


alicanerdurmaz (changes requested):
Hi, @damla thanks for the improving Refine!

There are a couple of missing things in the PR:

1. The `variant` prop was introduced in `antd@5.24.0`, but our `@refinedev/antd` package is using version `5.17.0`. That’s why our tests are failing. We need to update the `antd` dependency.

> You probably received this warning because you're using the latest version of `antd` in your own project.

2. If we’re going to make this update, it should also be applied to both [`<Edit />`](https://github.com/refinedev/refine/blob/main/packages/antd/src/components/crud/edit/index.tsx) and [`<Show />`](https://github.com/refinedev/refine/blob/main/packages/antd/src/components/crud/show/index.tsx) components.

3. We need a changeset to be able to release. You can find more details [here](https://refine.dev/docs/guides-concepts/contributing/#creating-a-changeset).

damla:
Oh okay then, I am closing this PR. Thanks for clarifying it 🙏🏼

Issue #6360, opened 2024-09-19
[BUG] Multiple Filters not Supported by Supabase Realtime


amerryma:
### Describe the bug

Right now, Supabase Realtime does not support multiple filters. If used, the live provider for supabase will send an invalid filter payload, causing the realtime request to fail.

### Steps To Reproduce

1. Create any realtime resource
2. Add more than 1 filter to the resource (I was using a useList hook with multiple objects in the filters option)
3. Watch websocket network requests, it will show an error.

### Expected behavior

I think we should either log a warning if we see multiple filters, or naively take the first one? 

### Packages

- supabase realtime v2.30.34 (not latest, but also not fixed in latest)
- @refinedev/supabase 5.9.4

### Additional Context

See here for Refine's implementation, where it is joining each filter with a comma. 

https://github.com/refinedev/refine/blame/8be95d808ae2f3f280ebd015bd95eb04238fe463/packages/supabase/src/liveProvider/index.ts#L67

See here for feature that is being discussed. No real activity it seems.

https://github.com/supabase/realtime/issues/486
https://github.com/supabase/realtime-js/issues/97

alicanerdurmaz:
Hello @amerryma, 

Can you please give us the `full error log` and `filters`?

alicanerdurmaz:
Hello @amerryma, thanks for the issue.

We discussed with the core team and decided not to add `filters` to the subscription since complex and relation-based filters might not work as expected in Supabase Realtime.

After removing `filters`, any change in the `posts` table will trigger the `posts` query again, ignoring the filters. We believe this approach is correct. What do you think?

amerryma:
My workaround for now was picking one filter that did 90% of the server side filtering then did the rest on the client side. So you're saying that you will remove the filtering altogether?

alicanerdurmaz:
> My workaround for now was picking one filter that did 90% of the server side filtering then did the rest on the client side. So you're saying that you will remove the filtering altogether?


I was wrong. [In this use case](https://github.com/refinedev/refine/discussions/6366), the filters must work. We'll investigate [this issue](https://github.com/supabase/realtime-js/issues/97) and see what we can do.

We're currently considering showing a warning when multiple filters are used and allowing users to override the `liveProvider`'s filters using the `meta.realtimeFilters` prop.

amerryma:
[...]

PR #6905, opened 2025-07-31
fix: support multiple filters in Supabase Realtime (#6360)

divyanshudg-git:
Fixes issue #6360

Updated mapFilter in liveProvider to handle multiple filters

Filters are now joined and passed correctly in Supabase subscription

Allows real-time updates with compound filter conditions

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,16386 @@
+{
+  "name": "refinedev",
+  "version": "0.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "refinedev",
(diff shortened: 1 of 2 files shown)


alicanerdurmaz (changes requested):
Hey @divyanshudg-git, Thank you very much for the PR!

I have a questions:

1. There’s no need to include `package-lock.json` — we’re using `pnpm`.
2. As far as I know, Supabase Realtime doesn’t support multiple filters. Has this changed? If so, could you please share a source?
3. Would you be able to add some tests?
4. Could you implement this feature in one of the examples to make it easier to test?

PR #7020, opened 2025-09-22
fix: use stable array to prevent memoization issue in useList #7019

### LINKED ISSUE #7019: [BUG]`useList`'s default `[]` return causes `useMemo` to break
### Describe the bug

In commit https://github.com/refinedev/refine/commit/43afc447b69a8d77bc3612db39dda1a5aa9baec0, the return value of the `useList` hook was changed. Specifically, `result.data` now defaults to a new empty array (`[]`) if `queryResponse.data.data` is falsy.

```typescript
// From: https://github.com/refinedev/refine/blob/acb1e96b0d6cfddbd6115818b8f01c73d56934ff/packages/core/src/hooks/data/useList.ts#L318

return {
  // ...
  result: {
    data: queryResponse?.data?.data || [],
    total: queryResponse?.data?.total,
  },
  // ...
};
```

While this change ensures that `result.data` is always an array, it introduces a significant side-effect: on every render cycle before the data is fetched, a **new instance** of an empty array is returned.

This breaks memoization with hooks like `useMemo` and `useEffect` because their dependency arrays rely on referential equality. Since `[] === []` is always `false`, any memoized calculation that depends on `result.data` will be re-executed on every render, leading to potential performance issues.

**Example:**

```tsx
const { data } = useList(...);

// This memoized value will be re-calculated on every render
// as long as `data` is the default empty array.
const memoizedData = useMemo(() => {
   // someComplexCompute(data)
   console.log("This will run on every render until data is loaded.");
   return someComplexCompute(data);
}, [data]);
```

### Steps To Reproduce

ses above

### Expected behavior

ses above

### Packages

- @refinedev/core

### Additional Context

_No response_

samay-rgb:
## PR Checklist

- [X] The commit message follows our guidelines: https://refine.dev/docs/guides-concepts/contributing/#commit-convention

## Bugs / Features

- [X] Related issue(s) linked

In commit 43afc44, the useList hook returns a new, empty array [] on every render cycle whenever queryResponse.data.data is falsy. This violates referential equality, causing hooks like useMemo and useEffect to re-calculate unnecessarily, leading to potential performance degradation.

## What is the new behavior?
The useList hook now returns a stable, frozen empty array Object.freeze([]) as the default value for result.data. 

fixes #7019


hongfanmeng:
The commit https://github.com/refinedev/refine/commit/43afc447b69a8d77bc3612db39dda1a5aa9baec0 also update `useTable` and `useMany`

packages/core/src/hooks/data/useList.ts
```diff
@@ -312,10 +312,12 @@ export const useList = <
     isLoading: queryResponse.isFetching,
   });
 
+  const EMPTY_ARRAY: readonly [] = Object.freeze([]);
```
alicanerdurmaz: I guess we need to initialize this outside the React component, otherwise it will be re-initialized on every render. or we must use `useMemo`

samay-rgb:
> Thanks for the improving Refine,
> 
> we need [changeset](https://refine.dev/docs/guides-concepts/contributing/#creating-a-changeset) to release this fix

Thanks for your feedback. I'll add the changeset.

packages/core/src/hooks/data/useList.ts
```diff
@@ -101,6 +101,8 @@ export type UseListReturnType<TData, TError> = {
   };
 } & UseLoadingOvertimeReturnType;
 
+const EMPTY_ARRAY: readonly [] = Object.freeze([]);
```
alicanerdurmaz, line 104: There is a type error here. Sorry, the example code I shared in the issue was wrong. I didn’t run it, so I didn’t notice.

```
Error: packages/core prepare: src/hooks/data/useList.ts(320,7): error TS2322: Type 'readonly [] | TData[]' is not assignable to type 'TData[]'.
packages/core prepare:   The type 'readonly []' is 'readonly' and cannot be assigned to the mutable type 'TData[]'.
Error: packages/core prepare: src/hooks/data/useMany.ts(264,7): error TS2322: Type 'readonly [] | TData[]' is not assignable to type 'TData[]'.
```

You can try removing `readonly []` or adding `as` to `EMPTY_ARRAY`. Removing `Object.freeze` might also fix the type error, since it’s already outside the component It may not re-render trigger but you should double check.

alicanerdurmaz (approved):

packages/core/src/hooks/data/useList.ts
alicanerdurmaz: There is a type error here. Sorry, the example code I shared in the issue was wrong. I didn’t run it, so I didn’t notice.

```
Error: packages/core prepare: src/hooks/data/useList.ts(320,7): error TS2322: Type 'readonly [] | TData[]' is not assignable to type 'TData[]'.
packages/core prepare:   The type 'readonly []' is 'readonly' and cannot be assigned to the mutable type 'TData[]'.
Error: packages/core prepare: src/hooks/data/useMany.ts(264,7): error TS2322: Type 'readonly [] | TData[]' is not assignable to type 'TData[]'.
```

You can try removing `readonly []` or adding `as` to `EMPTY_ARRAY`. Removing `Object.freeze` might also fix the type error, since it’s already outside the component It may not re-render trigger but you should double check.

alicanerdurmaz (changes requested):

alicanerdurmaz (approved):

PR #7030, opened 2025-09-29
fix: prevent external enabled from overriding internal ID check in useForm #7021

DipakHalkude:
## PR Checklist

- [ ] The commit message follows our guidelines: https://refine.dev/docs/guides-concepts/contributing/#commit-convention

## Bugs / Features

- [ ] Related issue(s) linked

## What is the new behavior?

fixes (issue)


DipakHalkude:
### Problem
The useForm hook was allowing external queryOptions.enabled property to override the internal ID safety check. This meant that even when id was undefined, the query would still execute if enabled was set to true externally, leading to unnecessary API calls.

### Solution
Reordered the property spreading in the useOne hook's queryOptions to ensure the internal enabled logic takes precedence over external enabled values.

Changes made:

- Modified packages/core/src/hooks/form/index.ts to move ...props.queryOptions before the enabled property
- Updated test in packages/core/src/hooks/form/index.spec.tsx to properly verify the fix
### Behavior
- Before : External enabled: true would override internal ID check, allowing queries with undefined IDs
- After : Internal ID check takes precedence - if id is undefined (and not a create action), query is disabled regardless of external enabled value
### Test Results
All 5 tests in the queryOptions.enabled behavior suite are now passing, confirming the fix works correctly.

DipakHalkude:
the Netlify failure appears to be environmental, not code-related

PR #7039, opened 2025-10-02
fix(core/useList): select function passed to useQuery in useList causes inifite loops

### LINKED ISSUE #6713: [BUG] `queryOptions.select` runs multiple times
### Describe the bug

I'm using GraphQL and I'm changing the data with the React Query's `select` middleware. It should run only once per request, but it is running multiple times. 

### Steps To Reproduce

1. Use the `useList` hook
2. Add `queryOptions.select: (data) => { console.log('select'); return data }`
3. See "select" is logged multiple times

Replication CodeSandbox: https://codesandbox.io/p/sandbox/react-query-5-forked-hh85pv

### Expected behavior

It should run only once. In React Query's documentation, they say to memoize the select function: https://tanstack.com/query/latest/docs/framework/react/guides/render-optimizations#memoization

`useList` is changing the function reference on every render.

### Packages

- "@refinedev/core": "^4.54.0",

### Additional Context

Possible fix: pass the `select` function inside a `useCallback` to `useList`'s `useQuery`

### LINKED ISSUE #7022: [BUG] setState in useList queryOptions select causes infinite loop
### Describe the bug

When using the `select` function in `queryOptions` of the `useList` hook, it causes an infinite re-rendering loop. The component continuously re-fetches data and never stabilizes, making the application unusable.

The issue occurs specifically when the `select` function is used to transform or filter the data returned from the API call within the `useList` hook's query options.

### Steps To Reproduce

1. Create a component that uses useList hook
2. Add queryOptions with a select function
3. Call setState or any state setter inside the select function
4. Observe infinite re-rendering in browser dev tools
 const { data } = useList({
  resource: "users",
  queryOptions: {
    select: (data) => {
      // This causes infinite loop
      setSomeState(data.items);
      return data;
    }
  }
});

Image

### Expected behavior

The select function should be able to handle side effects like setState calls without causing infinite re-renders. The query should execute once and allow state updates based on the transformed data without triggering additional query executions.

### Packages

 "@ant-design/icons": "^5.5.1",
    "@refinedev/antd": "^6.0.1",
    "@refinedev/cli": "^2.16.48",
    "@refinedev/core": "^5.0.1",
    "@refinedev/devtools": "^2.0.1",
    "@refinedev/react-router": "^2.0.0",

### Additional Context

This issue commonly occurs when developers need to update component state based on the transformed query data. The current behavior makes it impossible to perform side effects within the select function, limiting its usefulness for data transformation scenarios that require state updates.

filiperni:
## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

- Select function passed to useQuery is recreated on each render
- This causes unnecessary re-execution of the select function every time the useList hook runs.
- When the select function passed to useList (through queryProps) triggers a re-render, for example by updating some state, it causes a inifite loop. Like described in issue #7022.

## What is the new behavior?

fixes #7022 
fixes #6713

- Memoizes the select function passed to useQuery in useList hook
- Now select function is called only when:
  1. useQuery data changes
  2. pagination properties change in use useList

## Notes for reviewers

Added an example project to test out the behaviour of useList when passing query params with memoized select function.
I'm not sure if I should leave this project in repo or should I delete it. 

pnpm-lock file was automatically updated after running pnpm install


examples/use-list/src/App.css
```diff
@@ -0,0 +1,89 @@
...
+}
+
+.breadcrumb a {
+  color: blue;
+  text-decoration: none;
+}
+
+.menu {
+  flex-shrink: 0;
+  padding: 8px 16px;
```
maniya81: I think relative sizing would be better.

filiperni: Thank you for your comment. I would leave it as is because the UI in this example project is not important.
I created this example project using the `refine` CLI command to test out if the `useList` hook works properly.

filiperni: Actually, I don't even use these stylings. So I removed them.

filiperni:
@alicanerdurmaz can you take a look? When you'll have time :)

alicanerdurmaz:
Hey @filiperni , I just realized I made a mistake. I accidentally merged the newer [PR](https://github.com/refinedev/refine/pull/7106) instead of your earlier one ([#7039](https://github.com/refinedev/refine/pull/7039)).  
Your PR was really well done and a great improvement for Refine. I’m truly sorry for overlooking it.  

Thank you so much for your effort and for helping make Refine better!

Issue #7082, opened 2025-10-18
[BUG] useForm autoSave debounce leads to unbatched updates


kruschid:
### Describe the bug

Hi Refine team, thanks for the great library! I’ve noticed an issue with `useForm`. When using `useForm` from `@refinedev/core` with auto-save enabled, requests are triggered multiple times within the specified debounce window. This means the form state changes are not correctly grouped and debounced, instead, they are sent as individual delayed requests.

Looking at the source of the `useForm` hook, it appears that multiple instances of `asyncDebounce` are created. Each instance corresponds to a single form state update rather than grouping changes within the debounce window.

https://github.com/refinedev/refine/blob/8c403b2a890a0328c1991b40545d58bbb15e0fa9/packages/core/src/hooks/form/index.ts#L305-L309

I'm using headless refine with the latest version of Mantine. As a workaround, I wrapped the `onFinishAutoSave` function returned by `useFrom` in a `useDebouncedCallback` hook.

### Steps To Reproduce

```ts

import {  useForm } from "@refinedev/core";

useForm({
  autoSave: {
    enabled: true,
    debounce: 2000,
  },
  ...
});
```

Bind form-state changes to `onFinishAutoSave(values)` and modify any input field. Then observe how in the browser’s Network tab, multiple requests are triggered, even though the changes occur within the same time window of 2000ms.

### Expected behavior

Only one request should be sent after the user stops editing or when the debounce interval elapses. No multiple requests for changes occurring within the same window.

### Packages

@refinedev/core

PR #7083, opened 2025-10-19
fix(core): stabilize autosave debounce and cancel on unmount in useForm

a28689604:
- Memoize the asyncDebounce instance for `onFinishAutoSave` to avoid re-creating it on each render.
- Add cleanup to call `.cancel()` on unmount.
- Improves autosave reliability by preserving the debounce window across renders and avoiding stray timers.
- Fixes https://github.com/refinedev/refine/issues/7082

## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

A new debounced onFinishAutoSave is created on re-renders, which resets the debounce window and can cancel pending autosaves; in rapid re-renders (e.g., typing), autosave may not fire reliably. Pending timers aren’t explicitly cancelled on unmount.

## What is the new behavior?
onFinishAutoSave is memoized (one instance per debounce delay), invokes the latest onFinish via a ref, and cancels pending work on unmount. The debounce window is preserved across renders, making autosave reliable and preventing stray timers.

fixes [#7082 ](https://github.com/refinedev/refine/issues/7082)

## Notes for reviewers
This is my first open-source contribution to the project, happy to adjust anything based on your feedback.

### DIFF
--- a/packages/core/src/hooks/form/index.spec.tsx
+++ b/packages/core/src/hooks/form/index.spec.tsx
@@ -1347,4 +1347,138 @@ describe("useForm Hook", () => {
       });
     });
   });
+
+  describe("autoSave debounce stabilization", () => {
+    it("should preserve debounce window across re-renders and batch rapid changes", async () => {
+      vi.useFakeTimers();
+
+      const updateMock = vi
+        .fn()
+        .mockResolvedValue({ data: { id: "1", title: "test" } });
+
+      const { result, rerender } = renderHook(
+        ({ title }) =>
+          useForm({
+            resource: "posts",
+            action: "edit",
+            id: "1",
+            autoSave: {
+              enabled: true,
+              debounce: 1000,
+            },
+          }),
+        {
+          wrapper: TestWrapper({
+            dataProvider: {
+              default: {
+                ...MockJSONServer.default,
+                update: updateMock,
+              },
+            },
+            resources: [{ name: "posts" }],
+          }),
+          initialProps: { title: "initial" },
+        },
+      );
+
+      await act(async () => {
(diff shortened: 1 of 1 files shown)


packages/core/src/hooks/form/index.ts
```diff
@@ -302,12 +302,28 @@ export const useForm = <
...
   };
 
-  const onFinishAutoSave = asyncDebounce(
-    (values: TVariables) => onFinish(values, { isAutosave: true }),
-    props.autoSave?.debounce || 1000,
-    "Cancelled by debounce",
+  const onFinishRef = React.useRef(onFinish);
+  React.useEffect(() => {
+    onFinishRef.current = onFinish;
+  }, [onFinish]);
```
kruschid, line 308: Nice idea to prevent `asyncDebounce` from using a stale `onFinish`. But, if I'm not wrong, `useEffect` will be triggered on every (re-)render since `onFinish` is not memoized, so this could be simplified to `onFinishRef.current = onFinish` if you want to go with this pattern. What do you think?

a28689604: Thanks for the review!
I think it’s safer to keep the onFinishRef.current assignment inside a useEffect. This ensures the ref is only updated after commit, so it always reflects the latest committed onFinish.
If we assign it directly during render, we could risk writing an incorrect value in certain concurrent rendering scenarios, like when a render starts but later gets discarded. Wrapping it in useEffect avoids that and keeps the ref in sync with the committed tree.

kruschid: I wasn’t aware of that. Thanks for clarifying.

PR #7093, opened 2025-10-27
fix(core): correct useCustom return type to reflect empty object during loading

### LINKED ISSUE #7088: [BUG] `useCustom` return type is incorrect and undocumented v5 breaking change
### Describe the bug

Since this commit: 

```typescript
  result: {
  };
```

### Steps To Reproduce

```typescript
type MyCustomData = {
    id: string;
    someProperty: number[];
};

```

### Expected behavior

### Additional Context

_No response_

ghost:
## Summary
- Fixed the TypeScript return type for `useCustom` hook's `result.data` property
- The type now correctly reflects that `result.data` can be an empty object during loading

## Changes
- Updated `UseCustomReturnType` in `packages/core/src/hooks/data/useCustom.ts`
- Changed `data` type from `CustomResponse<TData>["data"]` to `CustomResponse<TData>["data"] | Record<string, never>`
- Added changeset documenting the fix

## Context
The `useCustom` hook implementation returns an empty object (`EMPTY_OBJECT`) for `result.data` while loading (line 236), but the TypeScript type definition didn't reflect this behavior. This caused TypeScript to not catch potential runtime errors when developers accessed properties on `result.data` during the loading state.

## Example
Before this fix, TypeScript would not error on the following code even though it would fail at runtime during loading:

```typescript
type MyCustomData = {
    id: string;
    someProperty: number[];
};

// TypeScript thinks someProperty is always defined, but it's not during loading
```

After this fix, TypeScript will correctly require checking if `result.data` has the expected properties before accessing them.

## Test Plan
- [ ] Verify that TypeScript now requires proper type guards when accessing properties on `result.data`
- [ ] Verify that existing code using `useCustom` with proper loading checks continues to work
- [ ] Confirm that the type accurately represents the runtime behavior

Resolves #7088

### DIFF
--- a/.changeset/fix-useCustom-return-type.md
+++ b/.changeset/fix-useCustom-return-type.md
@@ -1,0 +1,11 @@
+---
+"@refinedev/core": patch
+---
+
+fix(core): correct useCustom return type to reflect empty object during loading
+
+The `useCustom` hook returns an empty object for `result.data` while loading, but the TypeScript type only indicated it would be `CustomResponse<TData>["data"]`. This caused the type system to not catch potential runtime errors when accessing properties on `result.data` during the loading state.
+
+Updated the return type to `CustomResponse<TData>["data"] | Record<string, never>` to accurately represent that `result.data` can be an empty object while loading.
+
+[Resolves #7088](https://github.com/refinedev/refine/issues/7088)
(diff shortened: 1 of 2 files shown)


alicanerdurmaz:
> closed in favor of https://github.com/refinedev/refine/pull/7109

Hello @ctonneslan, I opened a new [PR](https://github.com/refinedev/refine/pull/7109) because I believe `useCustom` shouldn’t return an empty object, since we don’t know whether the API return value is an object, an array, or something else.  
I’m closing this PR for now. Feel free to give any advice or reopen it if you have any doubts.

PR #7100, opened 2025-11-03
fix(react-hook-form): resolve modal visibility issue when switching between actions

### LINKED ISSUE #7095: [BUG] `useModalForm` doesn't show after switching `action` from `edit` to `create`
### Describe the bug

I tried **Workaround 1** from https://github.com/refinedev/refine/issues/6451#issuecomment-2497667690 by @aliemir  and found that initially (i.e. using the initial `action` state) the modal does open, but opening it in `edit` mode and going back to `create` leaves `modal.visible` at `false` even though `modal.show()` was called.

### Steps To Reproduce

```tsx
import type { BaseKey, FormAction } from "@refinedev/core";
import { useModalForm } from "@refinedev/react-hook-form";
import { useCallback, useState } from "react";

{/*

    1. Click on "Open Create Modal"  ➝  State label will change to "visible" ✅

    2. Click on "Close Modal"  ➝  State label will change to "invisible" ✅

    3. Click on "Open Edit Modal"  ➝  State label will change to "visible" ✅

    4. Click on "Close Modal"  ➝ State label will change to "invisible" ✅

    5. Click on "Open Create Modal"  ➝ State label will stay at "invisible" ❌

/*}

export const Repro = () => {
	const [action, setAction] = useState<FormAction>();

	const form = useModalForm({
		refineCoreProps: {
			action: action,
			resource: "organization-contacts",
		},
	});

	const showModal = useCallback(
		(action: FormAction, id?: BaseKey) => {
			setAction(action);
			form.modal.show(id);
		},
		[form],
	);

	return (
		<div>
			<span>
				Current Modal State: {form.modal.visible ? "visible" : "invisible"}
			</span>
			<div
				style={{
					display: "flex",
					flexDirection: "row",
					gap: 16,
					alignItems: "center",
				}}
			>
				<button type="button" onClick={() => showModal("create")}>
					Open Create Modal
				</button>
				<button type="button" onClick={() => showModal("edit", 4)}>
					Open Edit Modal
				</button>
				<button type="button" onClick={() => form.modal.close()}>
					Close Modal
				</button>
			</div>
		</div>
	);
};
```

### Expected behavior

Switching from `action: "edit"` to `action: "create"` and calling `modal.show()` should make the `modal` visible.

### Packages

```
## System:
 - OS: Linux 6.16 Manjaro Linux
 - CPU: (16) x64 AMD Ryzen 7 7800X3D 8-Core Processor
## Binaries:
 - Node: 24.8.0 - /usr/bin/node
 - Yarn: 4.10.1 - /usr/bin/yarn
 - npm: 11.6.0 - /usr/bin/npm
## Browsers:
 - Chrome: Not Found
 - Firefox: 144.0
## Refine Packages:
 - @refinedev/core: 5.0.4
 - @refinedev/react-hook-form: 5.0.1
 - @refinedev/simple-rest: 6.0.0
```

AkprasadoP:
## Summary
Fixes a critical bug where `useModalForm` modal fails to show when switching from "edit" to "create" action.

## Root Cause
The `handleShow` callback in `useModalForm` had incomplete dependencies in its `useCallback`, causing stale closure issues when the `action` prop changed.

## Solution
- Added missing dependencies (`action`, `setId`, `show`) to the `useCallback`
- Added comprehensive test case reproducing the exact bug scenario

## Changes
- `packages/react-hook-form/src/useModalForm/index.ts`: Fixed useCallback dependencies
- `packages/react-hook-form/src/useModalForm/index.spec.ts`: Added test for action switching

## Testing
Added test case that reproduces the bug:
1. Open create modal ✅
2. Close modal ✅  
3. Switch to edit action and open modal ✅
4. Close modal ✅
5. Switch back to create action and open modal ✅ (was failing before fix)

Closes #7095

### DIFF
--- a/packages/react-hook-form/src/useModalForm/index.spec.ts
+++ b/packages/react-hook-form/src/useModalForm/index.spec.ts
@@ -438,7 +438,7 @@ describe("useModalForm Hook", () => {
     await act(async () => {
       result.current.modal.show();
     });
-    
+
     expect(result.current.modal.visible).toBe(true);
   });
 });


alicanerdurmaz (approved):

packages/react-hook-form/src/useModalForm/index.spec.ts
```diff
@@ -387,4 +387,59 @@ describe("useModalForm Hook", () => {
...
+    // Step 5: Close modal
+    await act(async () => {
+      result.current.modal.close();
+    });
+    expect(result.current.modal.visible).toBe(false);
+
+    // Step 6: Switch back to create action
+    rerender({ action: "create" });
+
+    // Step 7: Try to open create modal again - THIS IS WHERE THE BUG OCCURS
```
alicanerdurmaz: We shouldn’t add comments about issues to the codebase. It can be confusing. Beside that, everything is great! Thanks for improving Refine 🙌

AkprasadoP: Thanks for the feedback! 🙏
You're absolutely right about keeping the codebase clean. I've removed the bug reference comment from the test case.
Really appreciate you taking the time to review this — and thanks for the kind words about improving Refine!

PR #7106, opened 2025-11-11
fix(core): memoize queryOptions.select in useList hook

### LINKED ISSUE #6713: [BUG] `queryOptions.select` runs multiple times

angelodirb:
## Description

Fixes #6713 

This PR memoizes the `select` function in the `useList` hook to prevent it from being executed multiple times on each render.

## Problem

The `useList` hook was creating a new inline `select` function on every render, causing React Query to execute it multiple times unnecessarily. This is especially problematic when users provide their own `queryOptions.select` function.

## Solution

- Added `useMemo` to memoize the `select` function
- The function now maintains a stable reference across renders
- Only re-creates when dependencies (`prefferedPagination` or `queryOptions?.select`) change

## Changes

- Added `useMemo` import from React
- Wrapped the select logic in a memoized function
- Used the memoized version in the `useQuery` call

## Testing

- [x] Code compiles without errors
- [x] Lint-staged passes
- [x] Existing functionality preserved
- [x] Follows React Query best practices for memoization

## Related Documentation

- [React Query Memoization Guide](https://tanstack.com/query/latest/docs/framework/react/guides/render-optimizations#memoization)
- [Issue Discussion](https://github.com/refinedev/refine/issues/6713)

## Checklist

- [x] The commit message follows the convention
- [x] Related issue linked (#6713)
- [x] Changeset created
- [x] Code follows project style guidelines

### DIFF
--- a/.changeset/fix-modal-form-action-switching.md
+++ b/.changeset/fix-modal-form-action-switching.md
@@ -1,0 +1,5 @@
+---
+"@refinedev/react-hook-form": patch
+---
+
+Fix modal visibility issue when switching between actions in useModalForm hook. The modal now properly shows when switching from edit to create action.
--- a/.changeset/shiny-ways-burn.md
+++ b/.changeset/shiny-ways-burn.md
@@ -1,0 +1,7 @@
+---
+"@refinedev/mui": patch
+---
+
+fix: update TSDoc links in MUI button components
+
+Fixed Material UI documentation links in button components to point to the correct URL.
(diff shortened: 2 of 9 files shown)


clement-lta:
Sorry, could I understand why we use `useMemo` in this case instead of `useCallback` ?

PR #7109, opened 2025-11-11
fix(core): correct useCustom return type during loading

### LINKED ISSUE #7088: [BUG] `useCustom` return type is incorrect and undocumented v5 breaking change

alicanerdurmaz:
## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

The `useCustom` hook returns an empty object (`{}`) for `result.data` while loading, but the TypeScript type indicates it's `CustomResponse<TData>["data"]`. This causes TypeScript to miss runtime errors when accessing properties or methods on `result.data` without proper null checks.

For example, when expecting an array, calling `result.data.map()` fails at runtime with `result.data.map is not a function`.

## What is the new behavior?

Now `result.data` returns `undefined` while loading with the correct type `CustomResponse<TData>["data"] | undefined`. This forces developers to use proper null checks like `result.data?.map()`, preventing runtime errors.

fixes #7088

## Notes for reviewers

- Updated return type definition in `useCustom.ts`
- Removed `EMPTY_OBJECT` fallback in favor of returning `undefined`
- Added test case to verify `result.data` is `undefined` during loading state
- Updated existing test to check data is defined after successful load

### DIFF
--- a/.changeset/four-ads-sin.md
+++ b/.changeset/four-ads-sin.md
@@ -1,0 +1,9 @@
+---
+"@refinedev/core": patch
+---
+
+fix(core): correct `useCustom` return type to reflect undefined data during loading #7088
+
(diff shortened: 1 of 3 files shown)


alicanerdurmaz:
closed in favor of https://github.com/refinedev/refine/pull/7110

PR #7144, opened 2025-12-07
fix(core): correct useCustom return type to allow undefined data (fixes #7088)

### LINKED ISSUE #7088: [BUG] `useCustom` return type is incorrect and undocumented v5 breaking change

harikapadia999:
## Description

Fixes #7088 

The return type of `useCustom.result.data` was incorrectly typed as always present (`CustomResponse<TData>["data"]`), but the implementation could return `undefined` when `queryResponse.data` is undefined.

## Changes

- Updated `UseCustomReturnType` to make `result.data` optional: `TData | undefined`
- Removed the `EMPTY_OBJECT` fallback in favor of returning `undefined` directly
- Added JSDoc comment explaining that `data` can be `undefined`

## Breaking Change Note

This is technically a breaking change at the **type level only**, but it fixes incorrect types that were hiding potential runtime errors. The runtime behavior remains the same.

**Before:**
```typescript
const { result } = useCustom({ url: "/api/data", method: "get" });
// TypeScript thinks result.data is always TData
console.log(result.data.someProperty); // No type error, but could crash at runtime
```

**After:**
```typescript
const { result } = useCustom({ url: "/api/data", method: "get" });
// TypeScript correctly shows result.data can be undefined
console.log(result.data?.someProperty); // Type-safe with optional chaining
```

Users should add proper undefined checks when using `result.data`.

## Testing

- Existing tests pass
- Type checking now correctly identifies potential undefined access

## Checklist

- [x] Code follows project style guidelines
- [x] Changes are backward compatible (runtime behavior unchanged)
- [x] Documentation updated (JSDoc comments added)
- [x] Related issue linked (#7088)

### DIFF
--- a/packages/core/src/hooks/data/useCustom.ts
+++ b/packages/core/src/hooks/data/useCustom.ts
@@ -1,3 +1,4 @@
+import { useEffect } from "react";
 import { getXRay } from "@refinedev/devtools-internal";
 import {
   type QueryObserverResult,
@@ -5,11 +6,7 @@ import {
   useQuery,
 } from "@tanstack/react-query";
 
-import {
-  pickNotDeprecated,
-  useActiveAuthProvider,
-  prepareQueryContext,
-} from "@definitions/helpers";
+import { prepareQueryContext } from "@definitions/helpers";
 import {
   useDataProvider,
   useHandleNotification,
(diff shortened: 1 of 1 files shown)


BatuhanW:
Hey @harikapadia999 you checked out from master branch, please re-open it after checking out from `main` branch.

PR #7150, opened 2025-12-08
feat(core): add option to disable RouteChangeHandler

### LINKED ISSUE #7148: [FEAT] make RouteChangeHandler opt-out

Add a refine config option that will make `RouteChangeHandler` in the Refine container opt-out.

I have been using it, but the v5 upgrade has caused it to change behaviors as currently in my app it loads such the default object is setup for the context and it return false, but the checking is run outside of the Router context for me due to architectural reasons. So my app has historically had `RouteChangeHandler` be a no-op. So the simplest action is to just not render it in the tree.

_No response_

### Additional context

_No response_

```
                            <UnsavedWarnContextProvider>
                              <React.Fragment>
                                {children}
                                {!disableTelemetryWithDefault && <Telemetry />}
                                <RouteChangeHandler />
                              </React.Fragment>
                            </UnsavedWarnContextProvider>
```

to

```
                            <UnsavedWarnContextProvider>
                              <React.Fragment>
                                {children}
                                {!disableTelemetryWithDefault && <Telemetry />}
                                {!disableRouteChangeHandler && <RouteChangeHandler />}
                              </React.Fragment>
                            </UnsavedWarnContextProvider>
```

pcfreak30:
- Added `disableRouteChangeHandler` option to Refine context and props
- Implemented conditional rendering of RouteChangeHandler based on the new option
- Updated default options to disable the handler by default
- Added tests to verify the behavior with and without the handler
- Updated helper function to handle the new option

## PR Checklist

- [X] The commit message follows our guidelines: https://refine.dev/docs/guides-concepts/contributing/#commit-convention

## Bugs / Features

- [X] Related issue(s) linked
- [X] Tests for the changes have been added
- [X] Docs have been added / updated
- [X] Changesets have been added https://refine.dev/docs/guides-concepts/contributing/#creating-a-changeset

RouteChangeHandler always renders

## What is the new behavior?

RouteChangeHandler can be disabled with disableRouteChangeHandler.

fixes #7148

### DIFF
--- a/documentation/docs/core/refine-component/index.md
+++ b/documentation/docs/core/refine-component/index.md
@@ -475,7 +475,7 @@ Refine implements a simple and transparent telemetry module for collecting usage
 
 ### disableRouteChangeHandler
 
-`RouteChangeHandler` hooks into route changes to trigger navigation-side effects such as auth checks and redirect logic. Setting `disableRouteChangeHandler` to `true` disables these route-change side effects—not routing itself.
+`RouteChangeHandler` hooks into route changes to trigger navigation-side effects such as auth checks and redirect logic. Currently it invokes auth provider's check method and redirects to a different page if the user isn't authenticated. Setting `disableRouteChangeHandler` to `true` disables these route-change side effects—not routing itself.
 
 ### redirect
 


documentation/docs/core/refine-component/index.md
```diff
@@ -473,6 +473,10 @@ Refine implements a simple and transparent telemetry module for collecting usage
 
 > For more information, refer to the [Refine telemetry documentation &#8594](/further-readings/telemetry.md)
 
+### disableRouteChangeHandler
+
+`RouteChangeHandler` is a component that handles route changes within your Refine application. You can disable this by setting `disableRouteChangeHandler` to `true`.
+
```
erenkurnaz, line 479: Hello @pcfreak30, thanks for the contribution,
This description could be interpreted as “routing is being disabled.” If we update it similar to the one below, it would be clearer

>`RouteChangeHandler` hooks into route changes to trigger navigation-side effects such as auth checks and redirect logic. Setting `disableRouteChangeHandler` to true disables these route-change side effects—not routing itself.

pcfreak30: @erenkurnaz updated.

BatuhanW: Hey @pcfreak30 I think this could be more obvious for plain user, maybe mention something like, if you have an auth provider and check method returns false, means user is unauthenticated, this redirects user to another page, something like this.

BatuhanW: I think the main issue is about how we named this component, it's not really helping.

pcfreak30: @BatuhanW lmk what the docs should be then. im happy to use something you give as-is.

documentation/docs/core/refine-component/index.md
```diff
@@ -473,6 +473,10 @@ Refine implements a simple and transparent telemetry module for collecting usage
 
 > For more information, refer to the [Refine telemetry documentation &#8594](/further-readings/telemetry.md)
 
+### disableRouteChangeHandler
+
+`RouteChangeHandler` hooks into route changes to trigger navigation-side effects such as auth checks and redirect logic. Setting `disableRouteChangeHandler` to `true` disables these route-change side effects—not routing itself.
```
BatuhanW:
```suggestion
`RouteChangeHandler` hooks into route changes to trigger navigation-side effects such as auth checks and redirect logic. Currently it invokes auth provider's check method and redirects to a different page if the user isn't authenticated. Setting `disableRouteChangeHandler` to `true` disables these route-change side effects—not routing itself.
```

BatuhanW (approved):

BatuhanW:
Hello @pcfreak30 

Could you check our [module-federation](https://github.com/refinedev/refine/blob/main/examples/monorepo-module-federation/apps/host/src/App.tsx) example? I think it would be nice to add some config and/or comment highlighting case similar to yours.

pcfreak30:
> Hello @pcfreak30
> 
> Could you check our [module-federation](https://github.com/refinedev/refine/blob/main/examples/monorepo-module-federation/apps/host/src/App.tsx) example? I think it would be nice to add some config and/or comment highlighting case similar to yours.

Sure, but even in MF my case is almost abusing MF because I dont embed another app html. Its closer to the OG `require.js` that I use remotes as plugins that bring their own react components and I have a framework around it. So how I am using a lot of advanced things is as a unique, but complex electron-like app shell that loads defined plugins (from their `mf-manifest.json`) via a rest meta api.

So I don't follow any existing MF example :upside_down_face:

PR #7154, opened 2025-12-10
feat(ui): add ScrollToTop component

Biki-dev:
## Summary
Adds a reusable `ScrollToTop` component that appears after the user scrolls down and smoothly scrolls the page back to the top when clicked. The button is theme-aware and intended to be placed in the global layout (Root or Layout) so it works across all pages.

## Preview

https://github.com/user-attachments/assets/70ab3994-05f5-4e96-a1b0-715790a7a09c

## Features
- Appears after 300px of scrolling (configurable)
- Smooth scroll to top on click
- Fixed bottom-right position (configurable)
- Theme-aware: uses existing CSS variables / Tailwind classes to match site theme
- Accessible: keyboard focusable, aria-label, visible focus states
- Small, responsive animations and hover states

### DIFF
--- a/documentation/src/components/scroll-to-top/index.tsx
+++ b/documentation/src/components/scroll-to-top/index.tsx
@@ -1,0 +1,66 @@
+import React, { useState, useEffect } from "react";
+import clsx from "clsx";
+
+export const ScrollToTopButton: React.FC = () => {
+  const [isVisible, setIsVisible] = useState(false);
+
+  useEffect(() => {
+    const toggleVisibility = () => {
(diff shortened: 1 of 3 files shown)


Biki-dev:
This PR modifies documentation only (Docusaurus), so no changeset is required. Ready for review.
@BatuhanW @alicanerdurmaz

BatuhanW:
Hello @Biki-dev thanks for the PR! Unfortunately we can't merge it now, because we are changing the design completely. Once we release the new design, let's discuss it again and involve our designer too, so you can create a new PR.

PR #7161, opened 2025-12-16
fix(core): correct useCustom return type to allow undefined data (fixes #7088)

### LINKED ISSUE #7088: [BUG] `useCustom` return type is incorrect and undocumented v5 breaking change

harikapadia999:
## Description

Fixes #7088 

The `useCustom` hook's `result.data` type was incorrectly typed as always defined, causing TypeScript errors when using optional chaining or checking for undefined values.

### Type System
- Updated `UseCustomReturnType` to explicitly mark `result.data` as `TData | undefined`
- Added comprehensive JSDoc documentation explaining when data can be undefined
- Removed `EMPTY_OBJECT` fallback that masked the undefined state

### Runtime Behavior
- Now correctly returns `undefined` for `result.data` during:
  - Loading state (query is pending)
  - Error state (query failed)
  - Disabled queries (`enabled: false`)
  - Before first execution

### Testing
- Added 8 comprehensive test cases covering:
  - Undefined data during loading
  - Undefined data when query is disabled
  - Undefined data on error
  - Empty array vs undefined distinction
  - Null data handling
  - Optional chaining safety
  - State transition from undefined to defined

## Impact

✅ **Enables proper null-safety checks** - Users can now safely use optional chaining  
✅ **Aligns types with runtime** - TypeScript types now match actual behavior  
✅ **Backward compatible** - Existing code continues to work  
✅ **Better DX** - No more type errors for correct code  

## Example

**Before (Type Error):**
```typescript
const { result } = useCustom({ url: "/api/data", method: "get" });
const items = result.data?.items; // ❌ Type error: data is never undefined
```

**After (Works Correctly):**
```typescript
const { result } = useCustom({ url: "/api/data", method: "get" });
const items = result.data?.items; // ✅ Works! data can be undefined
```

### DIFF
--- a/.changeset/fix-usecustom-undefined-data.md
+++ b/.changeset/fix-usecustom-undefined-data.md
@@ -1,0 +1,20 @@
+---
+"@refinedev/core": patch
+---
+
+fix(core): correct useCustom return type to allow undefined data
+
+The `useCustom` hook's `result.data` type was incorrectly typed as always defined, causing TypeScript errors when trying to use optional chaining or check for undefined. This fix updates the return type to explicitly allow `undefined` for `result.data`, aligning TypeScript types with actual runtime behavior.
+
+**Changes:**
+- Updated `UseCustomReturnType` to mark `result.data` as `TData | undefined`
+- Added comprehensive JSDoc explaining when data can be undefined
+- Removed `EMPTY_OBJECT` fallback that masked the undefined state
+- Added extensive test coverage for undefined data scenarios
+
+**Impact:**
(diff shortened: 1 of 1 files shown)


hongfanmeng:
Has there been any progress on this? If you don’t plan to update it for now because this PR involves a breaking change, then at least fix the return type issue with `useCustom` first.

PR #7201, opened 2026-01-10
fix: supabase Realtime subscription to handle only the first filter

KajalMathe:
Issue:
Supabase Realtime currently supports only one filter per subscription. When multiple filters were provided, the live provider would send an invalid payload, causing the subscription to fail.

Fix:

Updated liveProvider in packages/supabase/src/liveProvider/index.ts.

Added mapFilter function to handle multiple filters by using only the first filter.

Added a console warning to inform users when multiple filters are provided.

Ensures that the filter parameter in Supabase Realtime is valid, preventing subscription errors.

Changes in this PR:

mapFilter function added to safely map filters.

Subscriptions now correctly apply only the first filter.

Console warning added for clarity.

All other functionality of liveProvider remains intact.

Example:

// Before: multiple filters would break subscription
subscribe({
  channel: "resources/posts",
  types: ["created", "updated"],
  params: {
    filters: [
      { field: "status", operator: "eq", value: "active" },
      { field: "category", operator: "eq", value: "tech" },
    ],
  },
  callback: () => {},
});

// After: only first filter is applied, subscription works

Testing:

Manual testing on local Supabase project confirmed that multiple filters no longer break the subscription.

Verified that single filters still work as expected.

Checked console warning appears when more than one filter is passed.

Related Issue:

(If the repo has an issue number, add it here, e.g., Fixes #1234)

Impact:

Improves stability for users subscribing with filters in Supabase Realtime.

Avoids runtime errors due to unsupported multiple filters.

### DIFF
--- a/packages/supabase/src/liveProvider/index.ts
+++ b/packages/supabase/src/liveProvider/index.ts
@@ -49,6 +49,8 @@ export const liveProvider = (
         }
       };
 
+      // TEMP: testing PR visibility
+
       // ✅ FIXED: Supabase Realtime supports only ONE filter
       const mapFilter = (filters?: CrudFilters): string | undefined => {
         if (!filters || filters.length === 0) {


alicanerdurmaz:
Hello @KajalMathe, thanks for the improving Refine but sorry, It's duplicate.

- closed in favor of [this PR](https://github.com/refinedev/refine/pull/7281).

PR #7266, opened 2026-02-11
fix(mui): prevent nested interactive elements in navigation buttons

### LINKED ISSUE #7085: [BUG] CreateButton of Material UI integration renders button in anchor
### Describe the bug

The [`CreateButton`](https://refine.dev/docs/ui-integrations/material-ui/components/buttons/create-button/#usage) of the Material UI integration renders a `button` element inside an `a` element:

Image

This is not recommended for accessibility reasons (see https://accessibleweb.com/question-answer/why-are-nested-interactive-controls-an-accessibility-issue/) and also harms keyboard usage: Tabbing through the application focuses the button twice (first the anchor then the button).

### Steps To Reproduce

1. Got to https://refine.dev/docs/ui-integrations/material-ui/components/buttons/create-button/#usage
2. Inspect the button
3. Test to focus the button with the keyboard -> It will be focused twice

### Expected behavior

Interactive elements should not be nested

### Packages

```
## System:
 - OS: macOS 15.5
 - CPU: (14) arm64 Apple M4 Pro
## Binaries:
 - Node: 24.5.0 - /opt/homebrew/bin/node
 - Yarn: 1.22.22 - /opt/homebrew/bin/yarn
 - npm: 11.5.1 - /opt/homebrew/bin/npm
## Browsers:
 - Chrome: 141.0.7390.108
 - Safari: 18.5
## Refine Packages:
 - @refinedev/cli: 2.16.48
 - @refinedev/core: 5.0.4
 - @refinedev/mui: 7.0.0
 - @refinedev/react-hook-form: 5.0.1
 - @refinedev/react-router: 2.0.1
```

### Additional Context

Relevant Refine code:
https://github.com/refinedev/refine/blob/8c403b2a890a0328c1991b40545d58bbb15e0fa9/packages/mui/src/components/buttons/create/index.tsx#L45-L76

kaigritun:
## Summary

This PR fixes an accessibility issue where navigation buttons (CreateButton, EditButton, ShowButton, ListButton, CloneButton) render a `<button>` element inside an `<a>` element.

## Problem

The current implementation wraps `<Button>` inside `<LinkComponent>`:

```jsx
<LinkComponent to={to}>
  <Button>...</Button>
</LinkComponent>
```

This creates nested interactive elements (`<a><button></a>`), which:
- Is not recommended for accessibility (see https://accessibleweb.com/question-answer/why-are-nested-interactive-controls-an-accessibility-issue/)
- Causes keyboard navigation issues (button is focused twice when tabbing)

## Solution

Use MUI's `component` prop to render the Button as the LinkComponent directly:

```jsx
<Button component={LinkComponent} to={to}>
  ...
</Button>
```

This approach:
- Maintains all existing button styling
- Preserves navigation behavior
- Fixes the nested interactive elements issue
- Follows MUI's recommended pattern for link buttons

## Changes

- `packages/mui/src/components/buttons/create/index.tsx`
- `packages/mui/src/components/buttons/edit/index.tsx`
- `packages/mui/src/components/buttons/show/index.tsx`
- `packages/mui/src/components/buttons/list/index.tsx`
- `packages/mui/src/components/buttons/clone/index.tsx`

## Testing

The fix maintains backward compatibility - all existing props and behaviors are preserved. The only change is the rendered HTML structure.

Fixes #7085


packages/mui/src/components/buttons/clone/index.tsx
```diff
@@ -44,9 +44,10 @@ export const CloneButton: React.FC<CloneButtonProps> = ({
   const { sx, ...restProps } = rest;
 
   return (
-    <LinkComponent
+    <Button
+      component={LinkComponent}
       to={to}
-      replace={false}
```
BatuhanW, line 49: Is removing replace intentional?

packages/mui/src/components/buttons/edit/index.tsx
```diff
@@ -56,27 +57,22 @@ export const EditButton: React.FC<EditButtonProps> = ({
           onClick(e);
         }
       }}
-      style={{ textDecoration: "none" }}
```
BatuhanW, line 59: The order of props are changed, so the diff is a bit hard to read. Could you try to keep current order as much as possible?

kaigritun:
Thanks for the review @BatuhanW! I've addressed both of your comments:

1. **replace prop**: Good catch! Removing `replace={false}` was not intentional. I've restored it in the latest commit. The prop is now passed to the Button component and will be forwarded to the LinkComponent via the `component` prop.

2. **Prop order**: I've reorganized the props to maintain the original order as much as possible. The new structure-specific props (component, to, replace) are added after `disabled` to minimize the diff, while keeping all other props in their original order (startIcon, title, sx, data-testid, className, ...restProps).

alicanerdurmaz (approved):

PR #7272, opened 2026-02-13
Fix: Supabase Realtime support for single filter with warning fallback

arpiiitver:
Currently, the Supabase Realtime provider crashes or errors out when multiple filters are provided. The `mapFilter` function attempts to join multiple filters with a comma (e.g., `filterA,filterB`), but the Supabase Realtime API expects a single filter string. This causes the WebSocket connection to fail or be rejected.

## What is the new behavior?
The provider now detects if multiple filters are present.
1.  It logs a `console.warn` to inform the developer that Supabase Realtime only supports a single filter.
2.  It explicitly selects only the first filter (`filters[0]`) to ensure a valid connection string is generated.
3.  This prevents the crash and allows the subscription to succeed with the primary filter.

## Notes for reviewers
I verified this fix locally by simulating a `subscribe` call with multiple filters.
- **Before fix:** The payload contained an invalid comma-separated string, causing errors.
- **After fix:** The payload correctly contained only the first filter, and the warning was logged to the console.

### DIFF
--- a/packages/supabase/src/liveProvider/index.ts
+++ b/packages/supabase/src/liveProvider/index.ts
@@ -54,17 +54,22 @@ export const liveProvider = (
           return;
         }
 
-        return filters
-          .map((filter: CrudFilter): string | undefined => {
-            if ("field" in filter) {
-              return `${filter.field}=${mapOperator(filter.operator)}.${
-                filter.value
-              }`;
-            }
-            return;
-          })
-          .filter(Boolean)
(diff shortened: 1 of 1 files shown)


BatuhanW:
Please make sure to thoroughly read and follow our contribution guide. PRs that doesn't follow our contribution guide, will be closed.

https://refine.dev/docs/guides-concepts/contributing/

PR #7355, opened 2026-03-20
fix(core): keep query context signal lazy when merging meta

### LINKED ISSUE #7132: [BUG] React Query double fetching in Refine v5 caused by unintended signal access (v4 did not do this)
### Describe the bug

Refine v5’s `useList` performs **two HTTP requests** in React 18 + Strict Mode in development, while refine v4 performs **only one**.

This is not caused by Strict Mode alone.
The root cause is that refine v5 **accidentally opts into React Query abort/cancel semantics** by *reading* the `context.signal` getter during `queryFn` execution via a spread operation:

```ts
const meta = {
  ...combinedMeta,
  ...prepareQueryContext(context), // <-- this spread triggers context.signal
};
```

Spreading an object with a `signal` getter **invokes the getter**, causing React Query to detect that `signal` is used. When `signal` is used, React Query treats the query as abortable, and in dev + Strict Mode it cancels the first run and issues a second fetch.

This behavior **did not happen in refine v4** because v4 passed the `signal` getter lazily (`meta.queryContext`) without accessing it inside `queryFn`.

This explains why:

* v4 = 1 request
* v5 = 2 requests
* Patching out the spread = v5 goes back to 1 request

### Steps To Reproduce

1. Create two minimal refine apps (one using refine v4, one using refine v5).
   My reproduction repo structure:
   ```
   .
   ├── refine-v4
   │   ├── src/App.jsx
   │   ├── index.html
   │   └── package.json
   └── refine-v5
       ├── src/App.jsx
       ├── index.html
       └── package.json
   ```
2. Both apps use the same simple `<useList>` component:
   ```jsx
   const { data, isLoading } = useList({ resource: "posts" });
   ```
3. Run both apps in React 18 with Strict Mode enabled (default Vite setup).
4. Open DevTools → Network tab
5. In **refine v4**, you will see **one** network request.
6. In **refine v5**, you will see **two** network requests.
7. Now patch refine v5’s compiled JS:
   Replace:
   ```js
   const meta2 = {
     ...combinedMeta,
     ...prepareQueryContext(context)
   };
   ```
   With:
   ```js
   const meta2 = {
     ...combinedMeta,
     queryKey: context.queryKey
   };

   Object.defineProperty(meta2, "signal", {
     enumerable: true,
     get() {
       return context.signal;
     },
   });
   ```
8. Restart refine v5. The second request disappears.

### Expected behavior

Refine v5 should behave like refine v4:

* **One request** in development + Strict Mode if the data provider does *not* explicitly opt in to abort controller semantics.
* `context.signal` should remain *lazy* and not be accessed inside `queryFn` until a data provider actually uses it.

### Packages

Refine v4:

```
yarn why @refinedev/core
└─ refine-v4-test@workspace:.
   └─ @refinedev/core@npm:4.58.0 [53edc] (via npm:^4.49.0 [53edc])
```

Refine v5:

```
yarn why @refinedev/core
└─ refine-v5-test@workspace:.
   └─ @refinedev/core@npm:5.0.6 [1e1cb] (via npm:^5.0.0 [1e1cb])
```

### Additional Context

* This behavior is fully explained by the TanStack Query team here:
  [https://github.com/TanStack/query/issues/3633](https://github.com/TanStack/query/issues/3633)
* Reading or destructuring `signal` marks the query as abortable.
* In refine v5, the spread operator `...prepareQueryContext(context)` **invokes the `signal` getter**, which implicitly accesses `context.signal`. That is the only difference from refine v4.
* Strict Mode double-render alone does **not** cause duplicate network requests unless the query is opt-in abortable.
* Two older refine issues referenced Strict Mode but did not identify the real cause:
  [https://github.com/refinedev/refine/issues/3449](https://github.com/refinedev/refine/issues/3449)
  [https://github.com/refinedev/refine/issues/2978](https://github.com/refinedev/refine/issues/2978)
* A minimal and backwards-compatible fix is simply avoiding the spread of an object with `signal` getter properties, and instead defining the lazy getter directly on `meta`, matching v4 behavior.

If needed, I can also provide a PR updating `useList`, `useOne`, `useMany`, etc., to ensure the `signal` getter is not invoked unless explicitly requested by the data provider.

Here is an example repo demonstrating the behavior: https://github.com/amerryma/refine-react-query-v4-v5-bug

Dailin521:
## Summary

This keeps `queryContext.signal` lazy when merging query context into `meta`.

Previously we spread `prepareQueryContext(context)` into a new `meta` object inside multiple core data hooks. Because `signal` is defined as a getter, object spread eagerly accessed it during merge. That marks the query as abortable earlier than intended and can trigger the duplicate request behavior described in #7132.

Closes #7132.

## Changes

- update `prepareQueryContext` to accept an optional `meta` object and merge into it internally
- preserve `signal` as a lazy getter on the returned object
- switch affected hooks to use the helper directly:
  - `useList`
  - `useOne`
  - `useMany`
  - `useInfiniteList`
  - `useCustom`
- add a regression test for lazy `signal` access

## Validation

- `corepack pnpm --filter @refinedev/devtools-shared --filter @refinedev/devtools-internal build`
- `corepack pnpm --dir packages/core test -- src/definitions/helpers/prepare-query-context/index.spec.ts src/hooks/data/useList.spec.tsx src/hooks/data/useOne.spec.tsx src/hooks/data/useMany.spec.tsx src/hooks/data/useCustom.spec.tsx src/hooks/data/useInfiniteList.spec.tsx`
- `corepack pnpm --dir packages/core build`

### DIFF
--- a/.changeset/chatty-cameras-remain.md
+++ b/.changeset/chatty-cameras-remain.md
@@ -1,0 +1,11 @@
+---
+"@refinedev/core": patch
+---
+
+fix(core): keep query context signal lazy when merging meta
+
+`prepareQueryContext` was previously spread into new `meta` objects inside core data hooks. Because `signal` is exposed as a getter, object spread eagerly accessed it during merge and marked the query as abortable earlier than intended.
+
+This keeps `signal` lazy by merging `meta` inside `prepareQueryContext` and preserving the getter on the returned object. The fix is applied across the affected hooks in `@refinedev/core`, and a regression test was added for the lazy `signal` behavior.
+
+Resolves #7132
--- a/packages/core/src/definitions/helpers/prepare-query-context/index.spec.ts
+++ b/packages/core/src/definitions/helpers/prepare-query-context/index.spec.ts
@@ -1,0 +1,23 @@
+import { prepareQueryContext } from ".";
+
+describe("prepareQueryContext", () => {
+  it("should keep signal lazy when merging it into meta", () => {
+    const signal = new AbortController().signal;
+    const signalGetter = vi.fn(() => signal);
+    const context = {
+      queryKey: ["posts", "list"],
+      get signal() {
+        return signalGetter();
+      },
+    } as any;
+
+    const meta = prepareQueryContext(context, { foo: "bar" });
+
+    expect(meta.foo).toBe("bar");
+    expect(meta.queryKey).toEqual(["posts", "list"]);
+    expect(signalGetter).not.toHaveBeenCalled();
+
+    expect(meta.signal).toBe(signal);
+    expect(signalGetter).toHaveBeenCalledTimes(1);
+  });
+});
--- a/packages/core/src/definitions/helpers/prepare-query-context/index.ts
+++ b/packages/core/src/definitions/helpers/prepare-query-context/index.ts
@@ -4,13 +4,19 @@ type Context =
   | QueryFunctionContext<QueryKey, any>
   | QueryFunctionContext<QueryKey, never>;
 
+type QueryContextMeta<TMeta extends Record<string, unknown> | undefined> = (
+  TMeta extends Record<string, unknown> ? TMeta : Record<string, never>
+) &
+  Pick<Context, "queryKey" | "signal">;
+
 export const prepareQueryContext = (
   context: Context,
-): Pick<Context, "queryKey" | "signal"> => {
-  const queryContext: Pick<Context, "queryKey" | "signal"> = {
+  meta?: Record<string, unknown>,
+): QueryContextMeta<typeof meta> => {
+  const queryContext = {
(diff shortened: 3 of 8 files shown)


Dailin521:
This PR is still intentionally narrow to the lazy `queryContext.signal` merge path in `@refinedev/core`.

Verification I ran is listed in the PR body and covers the targeted helper test, affected hook tests, and the package build. If there is any preferred adjustment to the approach or scope, I can update it quickly.

jwgmeligmeyling:
Change seems trivial. Any particular reason why the approach differs from the initial fix (that was unfortunately regressed) in #5851?

jwgmeligmeyling:
@Dailin521 why close the PR?

Issue #7198, opened 2026-01-09
[FEAT] Add fallback and namespace for useDocumentTitle


indramal:
### Is your feature request related to a problem? Please describe.

I can see when I add page title using `useDocumentTitle`, Sometimes, it is not available in i18n JSON file. So I suggest to use fallback/default title for page.

Also sometime I use different namespace files for each pages. I unable to set namespace for `useDocumentTitle`. 

I use below code when have `useDocumentTitle` hook.

```js
const translate= useTranslate();

  useEffect(() => {
    document.title = translate("pages.login.title", "Login") ;
  }, [t]);
```

### Describe the thing to improve

I can see same implementation use in `useDocumentTitle` hook - https://github.com/refinedev/refine/blob/51ad5ecb0dd7d5fcf91af94b6d4b9df7784de0cf/packages/react-router/src/use-document-title.ts#L6

So I suggest to add fallback/default title and namespace.

I suggest code:

```js
import { useTranslate } from "@refinedev/core";
import { useCallback, useEffect } from "react";

type Title = string | { i18nKey: string };

type UseDocumentTitleOptions = {
  namespace?: string;
  defaultTitle?: string;
};

export const useDocumentTitle = (
  title?: Title,
  options?: UseDocumentTitleOptions
) => {
  const translate = useTranslate();

  const getTitle = useCallback(
    (title: Title) => {
      const key = typeof title === "string" ? title : title.i18nKey;

      return translate(
        key,
        {
          ns: options?.namespace,
        },
        options?.defaultTitle
      );
    },
    [translate, options?.defaultTitle, options?.namespace]
  );

  useEffect(() => {
    if (!title) return;

    document.title = getTitle(title);
  }, [title, getTitle]);

  return (title: Title) => {
    document.title = getTitle(title);
  };
};
```

Manthancode28:
Hi @indramal 👋  
I’d like to work on this feature and submit a PR.  
Please assign this issue to me if it’s available. Thanks!

arndom:
not stale

PR #7386, opened 2026-04-08
feat(core): add namespace options to useTranslation

### LINKED ISSUE #7195: [FEAT] Loading namespaces with useTranslation and useTranslate hooks.

Current method we need to add namespaces for each `t`. 

According Refine docs, https://refine.dev/core/docs/i18n/i18n-provider/#translate

```js
const { translate } = useTranslation();
translate("posts.fields.title", { ns: "resources" }, "Title");
```

But i18next support directly load with `useTranslation` hook - https://react.i18next.com/latest/usetranslation-hook#loading-namespaces

```js
// load a specific namespace
// the t function will be set to that namespace as default
const { t, i18n } = useTranslation('ns1');
t('key'); // will be looked up from namespace ns1

// load multiple namespaces
// the t function will be set to first namespace as default
const { t, i18n } = useTranslation(['ns1', 'ns2', 'ns3']);
t('key'); // will be looked up from namespace ns1
```

This method will reduce boilerplate codes. 

_No response_

### Additional context

_No response_

I believe `useTranslation` and `useTranslate` get namespaces input as array or string like i18next package provide. Then directly give to options. Current method can not give namespaces as input.

https://github.com/refinedev/refine/blob/f73d69ff8c94303bff4b47f96227ab730caef574/packages/core/src/hooks/i18n/useTranslation.tsx#L14

https://github.com/refinedev/refine/blob/f73d69ff8c94303bff4b47f96227ab730caef574/packages/core/src/hooks/i18n/useTranslate.ts#L11

arndom:
## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

- namespace had to be specified for each key translation
- useDocumentTitle had no fallback

## What is the new behavior?
- all translations use the namespace from the hook call
- useDocumentTitle now has a fallback title

resolves #7195 #7198

## Notes for reviewers

Is anyone else facing this issue? I haven't contributed to testing since the vitest migration

image

### DIFF
--- a/packages/core/src/hooks/i18n/useTranslate.ts
+++ b/packages/core/src/hooks/i18n/useTranslate.ts
@@ -29,8 +29,13 @@ export const useTranslate = ({ ns }: UseTranslationProps = {}) => {
       let defaultMessage = _defaultMessage;
 
       if (typeof options === "string" && defaultMessage === undefined) {
-        options = undefined;
         defaultMessage = options;
+
+        if (ns) {
+          options = { ns };
+        } else {
+          options = undefined;
+        }
       }
 
       if (ns && options && typeof options !== "string") {
--- a/packages/core/src/hooks/i18n/useTranslate.ts
+++ b/packages/core/src/hooks/i18n/useTranslate.ts
@@ -1,6 +1,7 @@
 import { useContext, useMemo } from "react";
 
 import { I18nContext } from "@contexts/i18n";
+import type { UseTranslationProps } from "./useTranslation";
 
 /**
  * If you need to translate the texts in your own components, refine provides the `useTranslate` hook.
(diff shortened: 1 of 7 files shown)


packages/core/src/hooks/i18n/useTranslate.ts
```diff
@@ -21,14 +22,26 @@ export const useTranslate = () => {
...
-      options?: string | any,
-      defaultMessage?: string,
+      _options?: string | any,
+      _defaultMessage?: string,
     ) {
+      let options = _options;
+      let defaultMessage = _defaultMessage;
+
+      if (typeof options === "string" && defaultMessage === undefined) {
+        options = undefined;
```
sviat-barbutsa: I think this branch regresses the existing `translate(key, "fallback")` overload when `ns` is configured because `options` is set to `undefined` before assigning `defaultMessage` and in this case a call like: `translate("posts.fields.title", "Title")` will call `i18nProvider.translate` with both `options` and `defaultMessage` as `undefined`.

It also would not pass `{ ns: "resources" }`, because the namespace merge only runs when `options` is already an object.

Could you, please, add a regression test for this overload? 
I would expect  `useTranslate({ ns: "resources" })("posts.fields.title", "Title")` to call: `i18nProvider.translate("posts.fields.title", { ns: "resources" }, "Title")`

arndom: Thanks for pointing this out, didn't notice, will fix.

sviat-barbutsa: np, glad to help! Thank you

PR #7402, opened 2026-04-20
fix(react-hook-form): initialize useFieldArray via reset on first data load (fixes #7401)

### LINKED ISSUE #7401: [BUG] useFieldArray fields empty on second visit when using useForm
### Describe the bug

Potential fix:

```patch
```

### Steps To Reproduce

2. Go back to the list page

### Expected behavior

### Packages

- @refinedev/core: 5.0.12

### Additional Context

#### Workaround

```patch
```

#### Additional comments

Woutuuur:
Fixes #7401.

## PR Checklist

 - [X]  The commit message follows our guidelines: 
https://refine.dev/docs/guides-concepts/contributing/#commit-convention

## Bugs / Features

 - [X]  Related issue(s) linked
 - [X]  Tests for the changes have been added
 - [ ]  Docs have been added / updated
 - [X]  Changesets have been added https://refine.dev/docs/guides-concepts/contributing/#creating-a-changeset

When using `useFieldArray` inside a form powered by `useForm`, the fields array is empty on render even though the API response contains data. This happens consistentlyin cases where the component is retriggered when React Query has cached data. 

This seems to mostly affect cases where the `useFieldArray` render depends on `formLoading`, e.g.:

```tsx
const { refineCore: { formLoading } } = useForm(...);
{formLoading && <ComponentContainingUseFieldArray />}
```

See https://codesandbox.io/p/sandbox/t6pxft for a minimal replication.

## What is the new behavior?

`useFieldArray.fields` correctly reflects the array data from the API response.

## Notes for reviewers
 
`useFieldArray` must be mounted and initialized before `setValue` calls can populate it. With cached React Query data, the response is served immediately while also starting a background refetch, so `query.data` is already set when the component mounts, but `formLoading` is `true` (driven by `isFetching`). The [`[query.data]` effect](https://github.com/refinedev/refine/blob/main/packages/react-hook-form/src/useForm/index.ts#L288) fires right away, but if `useFieldArray` is behind a loading guard based on `formLoading` at that point, the `setValue` calls from [applyValuesToFields](https://github.com/refinedev/refine/blob/main/packages/react-hook-form/src/useForm/index.ts#L271) are a no-op. When `formLoading` turns `false`, `query.data` is the same object reference and the effect does not re-fire. In the end, `useFieldArray` mounts but `fields` stays empty.

The fix calls `reset({ ...getValues(), ...data }, { keepDirtyValues: true })` on the first data laod. Unlike `setValue`, `reset` sets the form's default values directly, so `useFieldArray` can initialize from them whenever. Merging with `getValues()` preserves any fields present in the form but absent from the API response. `keepDirtyValues: true` preserves values the user has already edited. Subsequent data loads use the existing `applyValuesToFields`.

### DIFF
--- a/packages/react-hook-form/src/useForm/index.spec.tsx
+++ b/packages/react-hook-form/src/useForm/index.spec.tsx
@@ -1,12 +1,12 @@
 import React from "react";
 
-import { useForm } from ".";
-import type { IRefineOptions, HttpError } from "@refinedev/core";
+import type { HttpError, IRefineOptions } from "@refinedev/core";
 import * as Core from "@refinedev/core";
-import { MockJSONServer, TestWrapper, act, render, waitFor } from "../../test";
-import { Route, Routes } from "react-router";
-import { Controller, useFieldArray } from "react-hook-form";
 import { screen } from "@testing-library/react";
+import { Controller, useFieldArray } from "react-hook-form";
+import { Route, Routes } from "react-router";
+import { useForm } from ".";
+import { MockJSONServer, TestWrapper, act, render, waitFor } from "../../test";
 
 interface IPost {
   title: string;
@@ -358,7 +358,7 @@ describe("useForm hook", () => {
 
     const EditPage = ({ formLoading }: { formLoading: boolean }) => {
       const { control } = useForm<IPost, HttpError, IPost>({
-        refineCoreProps: { resource: "posts", action: "edit", id: "1" }
+        refineCoreProps: { resource: "posts", action: "edit", id: "1" },
       });
       if (formLoading) return <p>loading</p>;
       return <FieldArrayChild control={control} />;
--- a/packages/react-hook-form/src/useForm/index.spec.tsx
+++ b/packages/react-hook-form/src/useForm/index.spec.tsx
@@ -5,7 +5,8 @@ import type { IRefineOptions, HttpError } from "@refinedev/core";
 import * as Core from "@refinedev/core";
 import { MockJSONServer, TestWrapper, act, render, waitFor } from "../../test";
 import { Route, Routes } from "react-router";
-import { Controller } from "react-hook-form";
+import { Controller, useFieldArray } from "react-hook-form";
+import { screen } from "@testing-library/react";
 
 interface IPost {
   title: string;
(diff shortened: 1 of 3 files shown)


jsingh6:
**Preflight flagged 2 potential issues worth a look:**

---
**`packages/react-hook-form/src/useForm/index.spec.tsx:357`** · `preflight/typescript/any-cast` · MEDIUM

Wondering if there could be an issue with `const FieldArrayChild = ({ control }: { control: any }) => {` here?

The `control` parameter is explicitly typed as `any`, disabling TypeScript's type checker for that value. It should use the proper `Control` type from react-hook-form instead.

> **Suggestion:** could this be restructured to avoid the problem described above?

```
const FieldArrayChild = ({ control }: { control: any }) => {
```

---
**`packages/react-hook-form/src/useForm/index.ts:277`** · `preflight/typescript/any-cast` · MEDIUM

Wondering if there could be an issue with `reset({ ...getValues(), ...data } as unknown as TVariables, …` here?

The value is cast through `unknown` to `TVariables` to silence type errors. While `unknown` is used as an intermediate step, the final cast to `TVariables` without a type guard means the actual shape is not verified, effectively bypassing type safety.

> **Suggestion:** could this be restructured to avoid the problem described above?

```
reset({ ...getValues(), ...data } as unknown as TVariables, {
```

---
*Posted by [Preflight](https://github.com/jsingh6/preflight)*

packages/react-hook-form/src/useForm/index.spec.tsx
```diff
@@ -329,4 +330,56 @@ describe("useForm hook", () => {
...
+    const FieldArrayChild = ({ control }: { control: any }) => {
+      const { fields } = useFieldArray({ control, name: "tags" });
+      return (
+        <ul>
+          {fields.map((field) => (
+            <li key={field.id} data-testid="tag-item" />
+          ))}
+        </ul>
+      );
+    };
```
jsingh6, line 357: **`preflight/typescript/any-cast` · MEDIUM**

Wondering if there could be an issue with `const FieldArrayChild = ({ control }: { control: any }) => {` here?

The `control` parameter is explicitly typed as `any`, disabling TypeScript's type checker for that value. It should use the proper `Control` type from react-hook-form instead.

> **Suggestion:** could this be restructured to avoid the problem described above?

*Posted by [Preflight](https://github.com/jsingh6/preflight)*

BatuhanW (approved):

sgmeligmeyling:
There's still an issue with this PR unfortunately where the form is now reset with all the fields of a record instead of only the registered fields. I've made a fix here: https://github.com/athenagroup/refine/commit/8ce02c4891637169a2240a13e4d2c82d4a5678ba

PR #7405, opened 2026-04-26
feat(core): add options.logout.confirm for logout confirmation dialog

Amaanghazi22:
## PR Checklist

## Bugs / Features
Related issue: N/A (UX improvement proposal)

- [ ] Related issue(s) linked

Clicking logout in the sider or header immediately logs the user out with no confirmation. Users can accidentally log out while navigating.

## What is the new behavior?
A new `logout.confirm` option can be set in `<Refine options={{ logout: { confirm: true } }}>`. When enabled, a confirmation dialog is shown before logout executes. Defaults to false — no breaking change.

## Notes for reviewers

This is a UX improvement. The option defaults to false so existing behavior is fully preserved. 

- antd sider uses Modal.confirm (consistent with Ant Design patterns)
- mui sider uses window.confirm (can be upgraded to a MUI dialog if preferred)
- The example app-crm-minimal demonstrates the feature end-to-end

Happy to add tests, docs, and a changeset if maintainers want to move forward.

### DIFF
--- a/packages/core/src/contexts/refine/types.ts
+++ b/packages/core/src/contexts/refine/types.ts
@@ -102,6 +102,16 @@ export interface IRefineOptions {
     icon?: React.ReactNode;
     text?: React.ReactNode;
   };
+  /**
+   * Configure logout behavior across all Refine UI packages.
+   */
+  logout?: {
+    /**
+     * When `true`, a confirmation dialog is shown before logout executes.
+     * @default false
+     */
+    confirm?: boolean;
+  };
 }
 
 export interface IRefineContextOptions {
@@ -124,6 +134,9 @@ export interface IRefineContextOptions {
   title: {
     icon?: React.ReactNode;
     text?: React.ReactNode;
+  };
+  logout: {
+    confirm: boolean;
   };
 }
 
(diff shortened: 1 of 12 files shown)


packages/core/src/contexts/refine/types.ts
```diff
@@ -102,6 +102,16 @@ export interface IRefineOptions {
     icon?: React.ReactNode;
     text?: React.ReactNode;
   };
+  /**
+   * Configure logout behavior across all Refine UI packages.
+   */
+  logout?: {
```
sviat-barbutsa, line 108: Overall this looks like a useful addition. However, I think it needs either broader UI package coverage or narrower wording for the core option.

The new `IRefineOptions.logout` comment says that it configures logout behavior "across all Refine UI packages", but this PR only wires `logout.confirm` into the Ant Design and MUI `ThemedSider` implementations. The default Chakra UI and Mantine siders also call `useLogout()` directly, so 
`<Refine options={{ logout: { confirm: true } }}>` 
would behave differently depending on which UI package the app uses.

Could you either add the same confirm handling to 
`packages/chakra-ui/src/components/themedLayout/sider/index.tsx` 
and `packages/mantine/src/components/themedLayout/sider/index.tsx`, 
or scope the option/docs to the UI packages implemented in this PR? 
Since these packages already use the shared `layoutSiderTests`, a shared logout-confirm test there would also help prevent this behavior from drifting between UI integrations.

PR #7406, opened 2026-05-02
fix: reset field array values on data refetch during SPA navigation

### LINKED ISSUE #7401: [BUG] useFieldArray fields empty on second visit when using useForm

aayushbaluni:
## Problem
When using `useForm` with `useFieldArray`, field arrays are populated correctly on first visit but empty on subsequent SPA navigations because the form state isn't properly refreshed with new query data.

## Change
Ensure the form is properly reset with fresh data when the query refetches on navigation, preserving field array population across SPA visits.

Fixes #7401.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/fix-field-array-spa-reset.md
+++ b/.changeset/fix-field-array-spa-reset.md
@@ -1,0 +1,5 @@
+---
+"@refinedev/react-hook-form": patch
+---
+
+Fix `useFieldArray` staying empty when query data is available before the field array mounts (e.g. cached React Query data with `formLoading`).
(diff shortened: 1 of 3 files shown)


Woutuuur:
Hi, there's already a PR open for this issue: https://github.com/refinedev/refine/pull/7402. Do you see any problems with its implementation? If so, let's discuss there instead of creating additional PRs!

PR #7408, opened 2026-05-03
Feat: enhance PostList UX with confirmations and improvements

Akkshita06:
- Added confirmation modal for bulk delete
- Added success and error notifications
- Added row-level delete button
- Optimized categoryIds using useMemo
- Improved loading and empty states


Akkshita06:
This change only improves example UX and does not affect any packages, so no changeset is required

PR #7417, opened 2026-05-11
fix(react-hook-form): useFieldArray fields empty on second visit

zons-zhaozhy:
Hey! Saw #7401 and figured I'd take a look.

The issue is that `applyValuesToFields(getRegisteredFields(), data, false)` misses `useFieldArray` fields when navigating back to the edit form. The problem: on the second visit, fieldArray fields register after the `queueMicrotask` callback runs, so `getRegisteredFields()` doesn't include them.

The fix replaces the manual `setValue` loop with RHF's `reset(data, { keepDirtyValues: true })`. `reset` handles field arrays natively and doesn't depend on field registration timing.

Changes:
- Added `reset` to the destructuring from `useHookFormResult`
- Replaced `applyValuesToFields(getRegisteredFields(), data, false)` with `reset(data, { keepDirtyValues: true })`

This matches the fix suggested in the issue. Let me know if anything needs tweaking.

### DIFF
--- a/packages/react-hook-form/src/useForm/index.ts
+++ b/packages/react-hook-form/src/useForm/index.ts
@@ -129,6 +129,7 @@ export const useForm = <
     watch,
     setValue,
     getValues,
+    reset,
     handleSubmit: handleSubmitReactHookForm,
     setError,
     formState: { dirtyFields },
@@ -268,7 +269,7 @@ export const useForm = <
     const applyQueryValues = () => {
       if (!isActive) return;
 
-      applyValuesToFields(getRegisteredFields(), data, false);
+      reset(data, { keepDirtyValues: true });
     };
 
     queryDataRef.current = data;


BatuhanW:
Hello @zons-zhaozhy please see if any additional tests are needed. Also you need to create changeset for us to release the fix. 

Please see details here: https://refine.dev/core/docs/guides-concepts/contributing/

Woutuuur:
Hi, there's already a PR open for this issue: https://github.com/refinedev/refine/pull/7402.  Do you see any problems with its implementation? If so, let's discuss there instead of creating additional PRs!

zons-zhaozhy:
Ah, I missed #7402 — that's a more complete fix with tests and changesets. Closing in favor of that one. Thanks @Woutuuur for pointing it out!

PR #7446, opened 2026-06-05
fix: hydrate field array values after form remount

### LINKED ISSUE #7401: [BUG] useFieldArray fields empty on second visit when using useForm

swayamshah01:
## Description

Fixes #7401.

This PR fixes an issue where `useFieldArray.fields` could be empty after navigating away from an edit form and then returning to it in an SPA.

The problem occurs when components using `useFieldArray` are rendered only after `useForm`'s `formLoading` becomes `false`. In that flow, the form receives the query data, but the field array may not be registered at the time values are applied. Hydrating fields with individual `setValue` calls does not reliably initialize React Hook Form's field array internals, so `fields` can remain empty even though the query response contains array data.

This updates query data hydration to use React Hook Form's `reset` API with `keepDirtyValues: true`. That allows field arrays to be initialized from the fetched record while preserving dirty user edits. The reset payload also merges the current form values with the query data so registered fields that are not present in the API response keep their existing/default values.

## Changes

- Replaced initial query-data hydration via registered-field `setValue` calls with `reset`.
- Preserved dirty values during hydration with `keepDirtyValues: true`.
- Added a regression test for navigating from an edit form to a list route and back to the edit form.
- Added a patch changeset for `@refinedev/react-hook-form`.

## Tests

```bash
pnpm.cmd --filter @refinedev/react-hook-form test
pnpm.cmd --filter @refinedev/react-hook-form build

### DIFF
--- a/.changeset/fix-use-field-array-navigation.md
+++ b/.changeset/fix-use-field-array-navigation.md
@@ -1,0 +1,5 @@
+---
+"@refinedev/react-hook-form": patch
+---
+
+Fixed query data hydration for forms using `useFieldArray` after SPA navigation.
--- a/packages/react-hook-form/src/useForm/index.spec.tsx
+++ b/packages/react-hook-form/src/useForm/index.spec.tsx
@@ -4,8 +4,8 @@ import { useForm } from ".";
 import type { IRefineOptions, HttpError } from "@refinedev/core";
 import * as Core from "@refinedev/core";
 import { MockJSONServer, TestWrapper, act, render, waitFor } from "../../test";
-import { Route, Routes } from "react-router";
-import { Controller } from "react-hook-form";
+import { Link, Route, Routes } from "react-router";
+import { Controller, type Control, useFieldArray } from "react-hook-form";
 
 interface IPost {
   title: string;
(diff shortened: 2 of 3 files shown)


Woutuuur:
Hi, there's already a PR open for this issue: https://github.com/refinedev/refine/pull/7402. Do you see any problems with its implementation? If so, let's discuss there instead of creating additional PRs!

swayamshah01:
Thanks for pointing me to #7402.

I checked its implementation and it solves the initial delayed `useFieldArray` registration case with `reset`, which is the same direction I took here.

One possible edge case I noticed is that #7402 only calls `reset` once via `hasResetRef`, then falls back to `setValue` for later query-data updates. Since `setValue` does not reliably update `useFieldArray.fields` metadata/length, I think field arrays may still get out of sync if the same mounted form receives new query data with a different array length.

For example:
- initial query data has `tags: [{ value: "react" }, { value: "refine" }]`
- later query data for the same mounted form has `tags: [{ value: "react" }]`
- after the first reset, the fallback `setValue` path may update values but leave `fields.length` as `2`

My PR resets on each query data change with `keepDirtyValues: true`, while merging current form values with query data to avoid dropping fields that are not present in the response.

Happy to move this discussion to #7402 if you prefer, or close this PR if maintainers decide #7402’s approach is sufficient.

PR #7456, opened 2026-06-15
fix(rest): type update.buildHeaders params with UpdateParams<any>

yashs33244:
## Summary

In `@refinedev/rest`, `UpdateParams<TVariables = {}>` defaults to `{}`. The `update.buildHeaders` option was typed `BuildHeaders<UpdateParams>` (no type arg), so a consumer's custom `update.buildHeaders` saw `params.variables` as `{}` and couldn't read any field off it without a TypeScript error.

This was the single outlier: all 9 other `buildHeaders` declarations, every other `update` option, and the default implementation use `<any>`. Sibling `update.buildBodyParams` / `transformError` already worked; only `update.buildHeaders` was affected.

## Changes

- `packages/rest/src/types.ts`: `BuildHeaders<UpdateParams>` → `BuildHeaders<UpdateParams<any>>`
- Add a `custom buildHeaders` test in `update.spec.ts` (reads `params.variables.foo`, asserts the derived header) + a `patch` changeset

## Testing

- Red-green proven against the real `@refinedev/core` types: reverting the line makes `update.spec.ts` fail with `TS2339: Property 'foo' does not exist on type '{}'`
- `biome ci` clean; full `@refinedev/rest` vitest suite 90 tests pass

### DIFF
--- a/.changeset/rest-update-buildheaders-variables-type.md
+++ b/.changeset/rest-update-buildheaders-variables-type.md
@@ -1,0 +1,7 @@
+---
+"@refinedev/rest": patch
+---
+
+fix(rest): type `update.buildHeaders` params with `UpdateParams<any>`
+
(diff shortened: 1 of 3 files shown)


yashs33244:
Hi @BatuhanW, would you mind taking a look when you have a moment?

This types `update.buildHeaders` params as `UpdateParams<any>` so a custom `buildHeaders` can actually read `params.variables` — it was the lone `buildHeaders` declaration still resolving to `{}` while every sibling option already used `<any>`. Includes a changeset and a test.

I've got two other small fixes open in the same spirit if they're useful: #7455 (typo fixes in user-facing error messages) and #7457 (reject non-finite values in `convertToNumberIfPossible`). Thanks!

PR #7475, opened 2026-06-30
fix(core): prevent cached show page data in create forms

Woutuuur:
## PR Checklist

## Bugs / Features

- [x] Related issue(s) linked

When opening a create modal (`useModalForm` with `action: "create"`) on a show page for the same resource, the form's `defaultValues` are overwritten by cached data from the show page's `useOne` query. This happens because `useForm` passes the URL-derived `id` to `useOne` even for create actions, causing a query key collision with the cached entry. The `useModalForm` visibility reset effect then calls `reset()` with the cached show-page data, replacing the user-provided `defaultValues`.

## What is the new behavior?

- core (`useForm`): `id` is no longer passed to `useOne` when the action is `"create"`, preventing the cache key collision at the source.
- react-hook-form (`useModalForm`): The visibility reset effect is guarded against `"create"` actions, so it never calls `reset()` with query data for create forms.

Create modal forms now correctly retain their `defaultValues` regardless of cached data from the same resource.

### DIFF
--- a/.changeset/fix-create-modal-cached-data.md
+++ b/.changeset/fix-create-modal-cached-data.md
@@ -1,0 +1,11 @@
+---
+"@refinedev/refine-core": patch
+"@refinedev/refine-react-hook-form": patch
+---
+
+fix(core, react-hook-form): prevent cached show-page data from overwriting create modal defaultValues
+
+When opening a create modal on a show page for the same resource, the form's `defaultValues` were overwritten by cached data from the show page's `useOne` query. This happened because `useForm` passed the URL-derived `id` to `useOne` even for create actions, causing a query key collision with the cached entry.
+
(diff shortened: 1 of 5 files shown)


abhishekchauhan1365:
This looks like a solid fix and the root cause is clearly addressed.

Avoiding passing the `id` to `useOne` for `"create"` actions makes sense and eliminates the cache key collision at the source. The additional safeguard in `useModalForm` to skip `reset()` for create flows is a good defensive layer as well.

The regression test is especially valuable here since this kind of issue can easily resurface with future changes.

One small thing to clarify:

* Are there any edge cases where `id` might still be implicitly inferred (e.g., via URL params in custom setups) that could reintroduce similar collisions?

Otherwise, the approach is clean and well-scoped. Good to go from my side.

Issue #7462, opened 2026-06-21
[BUG] MUI Breadcrumb links lose all styling — sx/underline/color/variant spread onto a plain <span>


Purav-Kanda:
### Describe the bug

In @refinedev/mui's Breadcrumb component, the internal LinkRouter helper is typed to accept MUI LinkProps (sx, underline, color, variant) but renders them onto a bare native <span> instead of an actual MUI component:

// packages/mui/src/components/breadcrumb/index.tsx
```
const LinkRouter = (props: LinkProps & { to?: string }) => {
  const { to, children, ...restProps } = props;
  return (
    <Link to={to || ""}>
      <span {...restProps}>{children}</span>
    </Link>
  );
};
```

Because <span> is a plain DOM element, none of the MUI styling props do anything. They just get dumped onto the DOM node as literal, invalid HTML attributes.

I confirmed this with a minimal repro (real React server render) using the exact call-site props this component passes:

```
<LinkRouter underline="hover" sx={{ display: "flex", alignItems: "center", fontSize: "14px", marginLeft: 0.5 }} color="inherit" to="/posts" variant="subtitle1">
  Posts
</LinkRouter>
```

Current (buggy) output:

`html<span underline="hover" sx="[object Object]" color="inherit" variant="subtitle1">Posts</span>
`
Visible impact: breadcrumb links aren't flex-aligned with their icons, get no hover underline, don't inherit color, and don't get the intended typography size.

Root cause: this is a regression from the v5 migration (#6945), specifically commit 5d63adac08 ("fix(mui): remove deprecated api"). Before that commit, LinkRouter correctly rendered MUI's actual Link:

```
const LinkRouter = (props: LinkProps & { to?: string }) => (
  <Link {...props} component={ActiveLink} />
);
```

When the legacy-router plumbing (useRouterContext, useRouterType, ActiveLink) was removed in favor of the single useLink() API, the component was rewritten to wrap a <span> — but the props meant for MUI's Link were never re-wired to an actual MUI component. The file still has an unused import { Link as MuiLink } from "@mui/material" (dead since that commit) — a leftover from an incomplete fix.

The shared breadcrumbTests suite (packages/ui-tests/src/tests/breadcrumb.tsx) only checks structure (text, href, icon presence), never styling, so this slipped through untested.

### Steps To Reproduce

1. Use @refinedev/mui's <Breadcrumb /> in any resource list/show/edit page with at least 2 breadcrumb levels (default minItems).
[...]

PR #7503, opened 2026-07-21
fix(mui): render Breadcrumb links as MUI Link so styling applies

adem-loghmari:
## PR Checklist

- [x] The commit message follows our guidelines
- [x] Related issue(s) linked
- [ ] Docs have been added / updated (not needed, no public API change)
- [x] Changesets have been added

In `@refinedev/mui`'s `Breadcrumb`, the internal `LinkRouter` helper is typed to accept MUI `LinkProps` (`sx`, `underline`, `color`, `variant`) but renders them onto a bare native `<span>`:

```tsx
const LinkRouter = (props: LinkProps & { to?: string }) => {
  const { to, children, ...restProps } = props;
  return (
    <Link to={to || ""}>
      <span {...restProps}>{children}</span>
    </Link>
  );
};
```

Because `<span>` is a plain DOM element, none of the MUI styling props do anything, they just get dumped onto the node as invalid HTML attributes. Visible impact: breadcrumb links aren't flex-aligned with their icons, get no hover underline, don't inherit color, and don't get the intended typography size.

This is a regression from the v5 migration (#6945), commit `5d63adac08`. Before that, `LinkRouter` rendered MUI's actual `Link`. When the legacy-router plumbing was removed in favor of `useLink()`, the component was rewritten to wrap a `<span>` but the props were never re-wired to a MUI component. The file still carried an unused `import { Link as MuiLink } from "@mui/material"` from that incomplete change.

fixes #7462

## What is the new behavior?

`LinkRouter` now renders a real MUI `Link`, using the router's link component as the underlying element:

```tsx
<MuiLink component={Link} to={to || ""} {...restProps}>
  {children}
</MuiLink>
```

This matches the existing pattern used by refine's MUI buttons (`component={LinkComponent}`), so the styling props (`sx`, `underline`, `color`, `variant`) are consumed by MUI while client-side routing through `to` is preserved. The previously-dead `MuiLink` import is now used.

## Notes for reviewers

The shared `breadcrumbTests` suite (`packages/ui-tests`) only checks structure (text, href, icon), so this slipped through untested. I added a MUI-specific test in `packages/mui/src/components/breadcrumb/index.spec.tsx` asserting the rendered link carries the `MuiLink-root` class and that no styling props leak into the DOM. Verified the test fails on the old `<span>` implementation and passes with the fix.

### DIFF
--- a/.changeset/mui-breadcrumb-link-styling.md
+++ b/.changeset/mui-breadcrumb-link-styling.md
@@ -1,0 +1,9 @@
+---
+"@refinedev/mui": patch
+---
+
+fix(mui): render `Breadcrumb` links as a MUI `Link` so styling is applied
+
+The internal `LinkRouter` helper in `@refinedev/mui`'s `Breadcrumb` spread MUI `Link` props (`sx`, `underline`, `color`, `variant`) onto a bare native `<span>`, so none of them had any effect. As a result breadcrumb links weren't flex-aligned with their icons, got no hover underline, didn't inherit color, and lost their typography size.
+
+`LinkRouter` now renders a real MUI `Link` (`component={LinkFromRouter}`), so the styling props are consumed by MUI while client-side routing via `to` is preserved. This also removes the previously-unused `Link as MuiLink` import that had been dead since the v5 migration.
--- a/packages/mui/src/components/breadcrumb/index.spec.tsx
+++ b/packages/mui/src/components/breadcrumb/index.spec.tsx
@@ -1,5 +1,8 @@
+import React from "react";
 import { vi } from "vitest";
 import { breadcrumbTests } from "@refinedev/ui-tests";
+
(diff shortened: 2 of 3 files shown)


adem-loghmari:
Not stale. Rebased onto the latest `main` (clean, no conflicts) and re-ran the package suite: all 8 tests in `packages/mui/src/components/breadcrumb` pass, including the styling test added here.

The bug is still reproducible on current `main`: `LinkRouter` still spreads the MUI `Link` props onto a plain `<span>`, so `sx`, `underline`, `color` and `variant` are silently dropped on every breadcrumb link. The `MuiLink` import on line 12 is still unused, which is the clearest sign this was an accident in the v5 migration (5d63adac08) rather than intent.

Happy to rebase again or split the test out if that helps review. Closes #7462.

PR #6515, opened 2024-11-21
fix: ensure CreateButton respects disabled prop in UI packages

OmkarBansod02:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://refine.dev/docs/guides-concepts/contributing/#commit-convention

## Bugs / Features

- [x] Related issue(s) linked
- [x] Tests for the changes have been added
- [ ] Docs have been added / updated
- [x] Changesets have been added https://refine.dev/docs/guides-concepts/contributing/#creating-a-changeset

## What is the current behavior?
  Create Button onClick still works when disabled prop is applied
## What is the new behavior?
  Create Button respects disabled prop in UI packages.

fixes #6513

PR #7518, opened 2026-08-06
fix: disabled not work on all navigation buttons

deepshekhardas:
fix


deepshekhardas:
Closing: superseded by upstream fix #6515 (ensure buttons respect disabled prop)

